PR #69, opened 2025-02-19
Add Cftlib + Config generator

banderous:
### Change description

Integrates the Cftlib + the CCD config generator

### To get started

Run `./gradlew bootWithCCD`

Starts PCS API + CCD & all dependencies

Once loaded open XUI at [http://localhost:3000](http://localhost:3000)

Login with user caseworker@pcs.com & blank pw (See CftlibConfig.java to see how this user is created)


JakeCG (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/CreateTestCase.java
```diff
@@ -0,0 +1,48 @@
...
+import uk.gov.hmcts.ccd.sdk.api.ConfigBuilder;
+import uk.gov.hmcts.ccd.sdk.api.Permission;
+import uk.gov.hmcts.ccd.sdk.api.callback.AboutToStartOrSubmitResponse;
+import uk.gov.hmcts.reform.pcs.ccd.domain.PCSCase;
+import uk.gov.hmcts.reform.pcs.ccd.domain.State;
+import uk.gov.hmcts.reform.pcs.ccd.domain.UserRole;
+
+@Profile("dev") // Non-prod event
+@Component
+public class CreateTestCase implements CCDConfig<PCSCase, State, UserRole> {
```
scottstewart-sl, line 16: Is there a way to avoid having this test data in the production code, (even if it is restricted by the profile)?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/UserRole.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+@AllArgsConstructor
+@Getter
+public enum UserRole implements HasRole {
+
+    CASE_WORKER("caseworker-civil", "CRU");
+
+    @JsonValue
+    private final String role;
+    private final String caseTypePermissions;
```
scottstewart-sl, line 19: The name suggests this could have multiple items - how would they be represented? If multiple permissions are actually possible then perhaps making this a Set or a List would be clearer?

peterpilgrim (approved):

MoLala1 (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/CreateTestCase.java
isha2040: This is being named like that, We can change it, it has nothing to do with testing. Its creating an event for the local CCD instance

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/UserRole.java
scottstewart-sl: In which case, what about using the enum `uk.gov.hmcts.ccd.sdk.api.Permission` which has those individual values, and also has a `toString()` method that will convert a `Set<Permission>` to a string with the letters in the correct order? It also has predefined `Set`s for `CR`, `CRU` and `CRUD`

That way it would avoid the risk of, say, a typo when entering the string value directly.

adusumillipraveen: Might be slightly an overkill given this is not user input, devs adding config in code. Would rather suggest setting a regex rule probably on the property to limit to something like 

`^[CRUD]{1,4}$` 
Lets see what Alex thinks :)

adusumillipraveen: There must be a validation on the service side i assume but yes, we can add on our side too.

peterpilgrim: I guess one would need a `Set<Permission>` type e.g. `HashSet<T>` ?

scottstewart-sl: It could just be a `Set<Permission>` as the field type and then use `Permission.CRU` in the constructor to specify the value.

It's not a biggie anyway, so I'm happy either way.

scottstewart-sl (approved):

PR #124, opened 2025-04-01
Add test data for post code look up

adusumillipraveen:
### Jira link

Adding test data as per https://tools.hmcts.net/jira/browse/HDPI-350

- Ideally this should be done by the data loader(but we do not have it for a bit now)
- This renames the preview migration file only on preview to a valid format and reverted to keep the workspace clean.
- File is ignored in other environments.

### Checklist

### DIFF
--- a/src/main/resources/db/migration/preview_test_data.sql
+++ b/src/main/resources/db/migration/preview_test_data.sql
@@ -1,0 +1,7 @@
+INSERT INTO postcode_court_mapping (postcode, epimid, legislative_country, effective_from, effective_to, audit) VALUES
+    ('W3 7RX', 20262, 'England', '', '2035-04-01', '{"created_by": "admin", "change_reason": "initial insert"}'::jsonb),
+    ('W3 6', 36791, 'England', '2025-02-01', '', '{"created_by": "admin", "change_reason": "initial insert"}'::jsonb),
+    ('M13 9PL', 144641, 'England', '2025-04-01', '2035-04-01', '{"created_by": "admin", "change_reason": "initial insert"}'::jsonb),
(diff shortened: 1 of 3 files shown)


MoLala1 (approved):

JakeCG (approved):

asifshaik-moj (approved):

src/main/resources/db/migration/preview_test_data.sql
```diff
@@ -0,0 +1,7 @@
+INSERT INTO postcode_court_mapping (postcode, epimid, legislative_country, effective_from, effective_to, audit) VALUES
+    ('W3 7RX', 20262, 'England', NULL, '2035-04-01 18:00:00', '{"created_by": "admin", "change_reason": "initial insert"}'::jsonb),
+    ('W3 6', 36791, 'England', '2025-02-01 08:00:00', NULL, '{"created_by": "admin", "change_reason": "initial insert"}'::jsonb),
```
TrevorHinsonHMCTS, line 3: W3 6RS - the RS seems to be missing from the test data in the Jira.

scottstewart-sl:
Instead of manipulating the migration file name, could you just have a separate resource directory for the test migrations and then change the flyway locations to include that extra directory on the preview environment? Then you would be keeping the test data out of the production code and you could also add more test data migration scripts in future if required.

adusumillipraveen:
> Instead of manipulating the migration file name, could you just have a separate resource directory for the test migrations and then change the flyway locations to include that extra directory on the preview environment? Then you would be keeping the test data out of the production code and you could also add more test data migration scripts in future if required.

yeah, i was mentioning in the standup that i wasn't happy with the approach though it works. I was trying the locations bit locally, so hopefully will have it working shortly.

adusumillipraveen:
Closing this in favour of #128

PR #142, opened 2025-04-14
HDPI 502 - add notifications to db

libanAbdirahman1:
### Jira link

See [HDPI-502](https://tools.hmcts.net/jira/browse/HDPI-502)

### Change description

Adding the Case Notification table to persist notification entities when a notification attempt is being made.  This will enable job scheduling and management of notifications when the endpoint is called in the future.

### Testing done

Added unit tests for new service method. Added integration tests for the controller. Also tested manually via postman if a notification entity was being saved when the endpoint was called successfully. 

### Checklist


src/integrationTest/java/uk/gov/hmcts/reform/pcs/notify/NotificationControllerIT.java
```diff
@@ -0,0 +1,155 @@
...
+        EmailNotificationRequest request = createEmailNotificationRequest();
+
+        mockMvc.perform(post(END_POINT)
+                            .contentType(MediaType.APPLICATION_JSON)
+                            .header(AUTHORIZATION, AUTH_HEADER)
+                            .header(SERVICE_AUTHORIZATION, SERVICE_AUTH_HEADER)
+                            .content(objectMapper.writeValueAsString(request)))
+                .andExpect(status().isOk());
+
+        verify(notificationClient).sendEmail(anyString(), anyString(), anyMap(), anyString());
```
scottstewart-sl, line 102: You could verify some of the parameters to `sendEmail` here

src/integrationTest/java/uk/gov/hmcts/reform/pcs/notify/NotificationControllerIT.java
```diff
@@ -0,0 +1,155 @@
...
+        .withPassword("testpass");
+
+    @DynamicPropertySource
+    static void overrideProperties(DynamicPropertyRegistry registry) {
+        registry.add("spring.datasource.url", postgres::getJdbcUrl);
+        registry.add("spring.datasource.username", postgres::getUsername);
+        registry.add("spring.datasource.password", postgres::getPassword);
+        registry.add("spring.jpa.hibernate.ddl-auto", () -> "update");
+        registry.add("spring.datasource.driver-class-name", postgres::getDriverClassName);
+    }
```
scottstewart-sl, line 65: There are a couple of PRs working their way through at the moment with slightly different ways of managing the Postgres test container. Let's review once they're all merged to `master` to make sure we are using the same approach for all the ITs that need a Postgres.

src/integrationTest/java/uk/gov/hmcts/reform/pcs/notify/NotificationControllerIT.java
```diff
@@ -0,0 +1,155 @@
...
+                            .header(AUTHORIZATION, AUTH_HEADER)
+                            .header(SERVICE_AUTHORIZATION, SERVICE_AUTH_HEADER)
+                            .content(objectMapper.writeValueAsString(request)))
+                .andExpect(status().isOk());
+
+        verify(notificationClient).sendEmail(anyString(), anyString(), anyMap(), anyString());
+    }
+
+    @Test
+    void testSavingNotificationWhenEndpointGetsCalled() throws Exception {
```
scottstewart-sl, line 106: Minor - I think we've agreed on a naming convention for tests that follows something along the lines of `shouldDoXWhenY`, with a @DisplayName annotation to add detail if necessary.

src/integrationTest/java/uk/gov/hmcts/reform/pcs/notify/NotificationControllerIT.java
```diff
@@ -0,0 +1,155 @@
...
+        EmailNotificationRequest request = createEmailNotificationRequest();
+
+        mockMvc.perform(post(END_POINT)
+                            .contentType(MediaType.APPLICATION_JSON)
+                            .header(AUTHORIZATION, AUTH_HEADER)
+                            .header(SERVICE_AUTHORIZATION,SERVICE_AUTH_HEADER)
```
scottstewart-sl, line 112: Are these 2 headers actually required? It might be that when using mockMvc then the auth filter is bypassed.

scottstewart-sl (commented):
Looks good - just a few small comments from me

src/integrationTest/java/uk/gov/hmcts/reform/pcs/notify/NotificationControllerIT.java
```diff
@@ -0,0 +1,155 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.springframework.http.HttpHeaders.AUTHORIZATION;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+import static uk.gov.hmcts.reform.pcs.hearings.constants.HearingConstants.SERVICE_AUTHORIZATION;
+
+@ExtendWith(SpringExtension.class)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
```
scottstewart-sl, line 46: Btw, you might be able to use WebEnvironment.MOCK (which is the default option) here if you are testing with `MockMvc`

src/integrationTest/java/uk/gov/hmcts/reform/pcs/notify/NotificationControllerIT.java
libanAbdirahman1: good catch, i've updated with the relevant parameters.

src/integrationTest/java/uk/gov/hmcts/reform/pcs/notify/NotificationControllerIT.java
libanAbdirahman1: more than happy with that naming convention and have updated accordingly.

src/integrationTest/java/uk/gov/hmcts/reform/pcs/notify/NotificationControllerIT.java
```diff
@@ -0,0 +1,149 @@
...
+
+        List<CaseNotification> notifications = notificationRepository.findAll();
+        assertThat(notifications).isNotEmpty();
+        assertThat(notifications.getFirst().getRecipient()).isEqualTo(request.getEmailAddress());
+        assertThat(notifications.getFirst().getStatus()).isEqualTo("Schedule Pending");
+    }
+
+    @DisplayName("Should return Internal Server Error when sending email fails")
+    @Test
+    void shouldReturnBadRequestWhenSendingEmailFails() throws Exception {
```
scottstewart-sl, line 114: Wouldn't a 502 error be better for this sort of thing? I tend to consider Internal Server Errors as unexpected errors that the developer hasn't caught.

If you do want to keep it as an Internal Server Error then the method test name needs updating.

src/integrationTest/java/uk/gov/hmcts/reform/pcs/notify/NotificationControllerIT.java
```diff
@@ -0,0 +1,149 @@
...
+import static org.mockito.Mockito.when;
+import static org.springframework.http.HttpHeaders.AUTHORIZATION;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+import static uk.gov.hmcts.reform.pcs.hearings.constants.HearingConstants.SERVICE_AUTHORIZATION;
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.MOCK)
+@AutoConfigureMockMvc
+@ActiveProfiles("integration")
+public class NotificationControllerIT extends AbstractPostgresContainerIT {
```
trevorhinson, line 42: Should this be called TestingSupportControllerIT instead?

src/main/resources/application.yaml
```diff
@@ -57,7 +57,7 @@ spring:
         show-sql: ${SPRING_JPA_SHOW_SQL:false}
     database-platform: org.hibernate.dialect.PostgreSQLDialect
   flyway:
-    enabled: false
+    enabled: true
```
trevorhinson, line 60: Is this to be turned on all the time now?   Suppose it makes sense to.

src/main/java/uk/gov/hmcts/reform/pcs/notify/domain/CaseNotification.java
```diff
@@ -0,0 +1,66 @@
...
+    private UUID caseId;
+
+    @Column(name = "submitted_at")
+    private LocalDateTime submittedAt;
+
+    @Column(name = "scheduled_at")
+    private LocalDateTime scheduledAt;
+
+    @Column(name = "last_updated_at",nullable = false)
+    private LocalDateTime lastUpdatedAt;
```
trevorhinson, line 46: Deviates from the Jira - called "lastUpdated"

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
```diff
@@ -93,4 +101,28 @@ public CompletableFuture<Notification> checkNotificationStatus(String notificati
             }
         });
     }
+
+    public CaseNotification createCaseNotification(String recipient, String type, UUID caseId) {
+        CaseNotification toSaveNotification = new CaseNotification();
+        toSaveNotification.setCaseId(caseId);
+        toSaveNotification.setStatus("Schedule Pending");
+        toSaveNotification.setType(type);
+        toSaveNotification.setRecipient(recipient);
```
trevorhinson, line 110: Is the last_updated not needed for the save?

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
```diff
@@ -93,4 +101,28 @@ public CompletableFuture<Notification> checkNotificationStatus(String notificati
             }
         });
     }
+
+    public CaseNotification createCaseNotification(String recipient, String type, UUID caseId) {
+        CaseNotification toSaveNotification = new CaseNotification();
+        toSaveNotification.setCaseId(caseId);
+        toSaveNotification.setStatus("Schedule Pending");
```
trevorhinson, line 108: Perhaps from an enum?

src/main/java/uk/gov/hmcts/reform/pcs/notify/domain/CaseNotification.java
```diff
@@ -0,0 +1,66 @@
...
+import lombok.Setter;
+import lombok.ToString;
+
+import java.time.LocalDateTime;
+import java.util.UUID;
+
+@Entity
+@AllArgsConstructor
+@NoArgsConstructor
+@Table(name = "case_notification")
```
trevorhinson, line 24: table is plural in the Jira

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
trevorhinson: Ah Ive seen the prepersist now.   Ignore this.

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
```diff
@@ -93,4 +101,28 @@ public CompletableFuture<Notification> checkNotificationStatus(String notificati
             }
         });
     }
+
+    public CaseNotification createCaseNotification(String recipient, String type, UUID caseId) {
```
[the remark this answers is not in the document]
jaina-sol: This is currently a testing support controller and would not be used in production. So validation will not be required here. Thanks.

src/main/resources/application.yaml
jaina-sol: Yes, we will leave this on now, as it works fine.

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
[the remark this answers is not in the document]
jaina-sol: It is public because it needs to be visible to the controller for testing. I didn't see a visible for testing. let me know if you think otherwise.

src/integrationTest/java/uk/gov/hmcts/reform/pcs/notify/NotificationControllerIT.java
jaina-sol: It was agreed with @adusumillipraveen to leave as it is, since it is just for testing and won't be used in production.

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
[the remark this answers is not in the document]
scottstewart-sl: From what I can see this method is only called from the unit test and not the controller. The unit test is in the same package so this could be package private as Trevor suggested.

However, I think the unit tests should be calling the public method `sendEmail` instead rather than exposing a method just to make writing the unit tests simpler. (In my opinion that's only something to be considered for legacy code). There is already a unit test which has some stubbing to call the `sendEmail` method, so you could base your new unit test stubbing on that and then when all is green refactor our the common stubbing code to keep the tests readable.

src/main/java/uk/gov/hmcts/reform/pcs/notify/domain/CaseNotification.java
```diff
@@ -0,0 +1,66 @@
...
+    private LocalDateTime lastUpdatedAt;
+
+    @Column(name = "status", nullable = false)
+    private String status;
+
+    @Column(name = "type", nullable = false)
+    private String type;
+
+    @Column(name = "recipient", nullable = false)
+    private String recipient;
```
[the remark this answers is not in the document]
jaina-sol: It was agreed to change the schema for the notification table. I will update the jira to reflect this. the only thing that remains as in Jira is the table name.

src/main/java/uk/gov/hmcts/reform/pcs/notify/domain/CaseNotification.java
jaina-sol: It was agreed to change the schema for the notification table. I will update the jira to reflect this. the only thing that remains as in Jira is the table name.

src/main/java/uk/gov/hmcts/reform/pcs/notify/domain/CaseNotification.java
```diff
@@ -0,0 +1,66 @@
...
+import lombok.Setter;
+import lombok.ToString;
+
+import java.time.LocalDateTime;
+import java.util.UUID;
+
+@Entity
+@AllArgsConstructor
+@NoArgsConstructor
+@Table(name = "case_notifications")
```
adusumillipraveen, line 24: Actually, just checking this. `case_notification` should be good for consistency.

src/main/java/uk/gov/hmcts/reform/pcs/notify/model/NotificationStatus.java
```diff
@@ -0,0 +1,36 @@
...
+
+public enum NotificationStatus {
+    CREATED("created"),
+    SENDING("sending"),
+    DELIVERED("delivered"),
+    PERMANENT_FAILURE("permanent-failure"),
+    TEMPORARY_FAILURE("temporary-failure"),
+    TECHNICAL_FAILURE("technical-failure"),
+    SCHEDULE("schedule"),
+    SCHEDULE_PENDING("schedule pending");
```
adusumillipraveen, line 11: Happy for this to be
```suggestion
    SCHEDULE_PENDING("schedule-pending");
```

scottstewart-sl (dismissed):

scottstewart-sl (approved):

src/main/java/uk/gov/hmcts/reform/pcs/notify/model/NotificationStatus.java
```diff
@@ -0,0 +1,36 @@
...
+    public String getValue() {
+        return value;
+    }
+
+    @Override
+    public String toString() {
+        return value;
+    }
+
+    public static NotificationStatus fromString(String status) {
```
trevorhinson, line 28: Was looking for test coverage but I do not believe this is use?

src/main/java/uk/gov/hmcts/reform/pcs/notify/domain/CaseNotification.java
```diff
@@ -0,0 +1,66 @@
...
+
+@Entity
+@AllArgsConstructor
+@NoArgsConstructor
+@Table(name = "case_notification")
+@Getter
+@Setter
+@ToString
+@EqualsAndHashCode(of = "notificationId")
+public class CaseNotification {
```
trevorhinson, line 29: Jira mentions a "referenceID" ?   Perhaps the Jira needs updating?

jaina-sol: @adusumillipraveen any thoughts on this, please?

adusumillipraveen: yes please, can we name it `provider_notification_id` ? this is the unique ID on the notification provider system (got notify)

adusumillipraveen: You dont need to use it now as it will be used by future tickets.

jaina-sol: Yes we have it already it is called `notification_id`
[https://github.com/hmcts/pcs-api/blob/68d0886fede9989fc5a47b51f908fd84d19949c6/src/main/resources/db/migration/V002__case_notification.sql#L3](url)

adusumillipraveen: that's an ID on our side, we need an external system ID to get the status. We need both

src/integrationTest/java/uk/gov/hmcts/reform/pcs/notify/NotificationControllerIT.java
jaina-sol: I have renamed it accordingly.

scottstewart-sl (approved):

scottstewart-sl (approved):

adusumillipraveen (approved):

src/integrationTest/java/uk/gov/hmcts/reform/pcs/notify/NotificationControllerIT.java
TrevorHinsonHMCTS: Sorry was this renamed?   Wondering if it is in the wrong package if is.

src/main/java/uk/gov/hmcts/reform/pcs/testingsupport/endpoint/TestingSupportExceptionHandler.java
```diff
@@ -0,0 +1,19 @@
...
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import uk.gov.hmcts.reform.pcs.controllers.RestExceptionHandler;
+import uk.gov.hmcts.reform.pcs.notify.exception.NotificationException;
+
+@ControllerAdvice(basePackages = "uk.gov.hmcts.reform.pcs.testingsupport")
+public class TestingSupportExceptionHandler {
+
+    @ExceptionHandler(NotificationException.class)
+    public ResponseEntity<RestExceptionHandler.Error> handleNotificationException(NotificationException ex) {
```
trevorhinson, line 14: Probably needs a test case associated with it.

MoLala1 (approved):

src/main/java/uk/gov/hmcts/reform/pcs/testingsupport/endpoint/TestingSupportExceptionHandler.java
```diff
@@ -0,0 +1,19 @@
...
+        return ResponseEntity
+            .status(HttpStatus.INTERNAL_SERVER_ERROR)
+            .body(new RestExceptionHandler.Error(ex.getMessage()));
```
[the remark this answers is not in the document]
MoLala1: dont see this being a security issue as we are only returning error message as this is just setting a status and returning the error to user.

isha2040 (approved):

src/main/resources/application.yaml
scottstewart-sl: There was presumably a reason why this had a default value of false in the first place, otherwise the property could have just been left out. I know that Jenkins applies the Flyway migrations as part of the pipeline in the deployed environments, so maybe it is safer to leave this as false and just provide an environment variable to override this locally. `SPRING_FLYWAY_ENABLED=true`

@adusumillipraveen - any thoughts?

src/test/java/uk/gov/hmcts/reform/pcs/testingsupport/endpoint/TestingSupportExceptionHandlerTest.java
```diff
@@ -0,0 +1,30 @@
...
+    private final TestingSupportExceptionHandler handler = new TestingSupportExceptionHandler();
+
+    @Test
+    void shouldReturnInternalServerErrorForNotificationException() {
+        NotificationException exception = mock(NotificationException.class);
+        when(exception.getMessage()).thenReturn("Failed to send notification");
+        
+        ResponseEntity<RestExceptionHandler.Error> response = handler.handleNotificationException(exception);
+        
+        assertEquals(HttpStatus.INTERNAL_SERVER_ERROR, response.getStatusCode());
```
scottstewart-sl, line 25: Minor - AssertJ is preferred over the Junit assertions

scottstewart-sl: This can be fixed in a future PR

src/main/resources/application.yaml
jaina-sol: I added my previous comment after discussing with Praveen, he's asked for us to leave it on.

scottstewart-sl (approved):

MoLala1 (approved):

src/main/resources/application.yaml
adusumillipraveen: I have originally kept it as optional as it was mostly for a specific use case and most of everything would work without a DB. But, i think with decentralised CCD it makes sense to have it default now. By default any instance running on kubernetes are configured to run in no-op https://github.com/hmcts/pcs-api/blob/master/charts/pcs-api/values.yaml#L26. So jenkins takes care of migration and app just validates if schema is up to date. So yes, no issues.

PR #186, opened 2025-05-06
HDPI-505: Write notification status to DB

andrewdami11:
### Jira link

See [HDPI-505](https://tools.hmcts.net/jira/browse/HDPI-505)

### Change description

- Write notification status in the DB, after [asynchronously] fetching it
- Fixed an issue where the `NotificationService` was trying to use a method that doesn't exist. The code was calling `getCaseId()` on `EmailNotificationRequest` objects, but this method isn't available. 
I updated the service to generate a random UUID for `caseIDs`, instead of trying to get one from the request. This ensures notifications can be created successfully, without changing the request model.

- I also fixed several tests that were failing because they expected different behavior than what the code actually does.

### Testing done

- Fixed the integration test to expect "sending" status instead of "pending-schedule"
- Updated unit tests to correctly verify the database save operations happen twice
- Added a new test to specifically check that UUIDs are correctly generated
- Fixed style warnings in the test code
- Verified all tests now pass successfully
- Checked database logs to confirm notifications are saved with correct status values (screenshot below)

### Screenshot

image

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
```diff
@@ -66,6 +83,15 @@ public SendEmailResponse sendEmail(EmailNotificationRequest emailRequest) {
...
             return sendEmailResponse;
         } catch (NotificationClientException notificationClientException) {
+            // Update notification status to failure
+            try {
+                caseNotification.setStatus(NotificationStatus.TECHNICAL_FAILURE.toString());
+                caseNotification.setLastUpdatedAt(LocalDateTime.now());
+                notificationRepository.save(caseNotification);
+            } catch (Exception e) {
+                log.error("Error updating notification status to failure: {}", e.getMessage(), e);
+            }
```
scottstewart-sl, line 93: You could extract this to a method called something like `updateNotificationStatusToFailure()` or `updateNotificationStatus(NotificationStatus)`. It would remove the need to have the code comment here explaining what the lines are doing and the latter option might also be re-usable by the code below that sets the status based on what comes back from the Notify API.

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
```diff
@@ -84,10 +110,33 @@ public CompletableFuture<Notification> checkNotificationStatus(String notificati
...
+                // Find and update notification status in database
+                try {
+                    CaseNotification caseNotification = notificationRepository
+                        .findByProviderNotificationId(UUID.fromString(notificationId))
+                        .orElse(null);
+                    
+                    if (caseNotification != null) {
+                        try {
+                            NotificationStatus notificationStatus = NotificationStatus.fromString(status);
+                            caseNotification.setStatus(notificationStatus.toString());
```
scottstewart-sl, line 124: Just out of interest, why isn't `CaseNotification` using `NotificationStatus` for the type of the status field? It would save having to read it `fromString` and then convert it back again with `toString()`.

src/test/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationServiceTest.java
```diff
@@ -102,7 +104,8 @@ void testSendEmailFailure() throws NotificationClientException {
             .hasMessage("Email failed to send, please try again.");
 
         verify(notificationClient).sendEmail(anyString(), anyString(), anyMap(), anyString());
-        verify(notificationRepository).save(any(CaseNotification.class));
+        // Verify save is called exactly 2 times - once for initial creation and once for status update
+        verify(notificationRepository, org.mockito.Mockito.times(2)).save(any(CaseNotification.class));
```
scottstewart-sl, line 108: Minor - can you do a static import for `times()` to make this code a bit neater? (Same in the test above)

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
```diff
@@ -55,6 +60,18 @@ public SendEmailResponse sendEmail(EmailNotificationRequest emailRequest) {
                 referenceId
             );
 
+            // Update notification with provider ID received from GOV.UK Notify
+            try {
+                UUID providerNotificationId = UUID.fromString(sendEmailResponse.getNotificationId().toString());
```
trevorhinson, line 65: Similar to what Scott has mentioned on the chunk below.   I would extract this out to reduce the cyclomatic complexity in this method and make this independently testable.

src/test/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationServiceTest.java
```diff
@@ -80,7 +81,8 @@ void testSendEmailSuccess() throws NotificationClientException {
             .isEqualTo(UUID.fromString("550e8400-e29b-41d4-a716-446655440000"));
         assertThat(response.getReference()).contains("reference");
         verify(notificationClient).sendEmail(anyString(), anyString(), anyMap(), anyString());
-        verify(notificationRepository).save(any(CaseNotification.class));
+        // Verify save is called exactly 2 times - once for initial creation and once for status update
+        verify(notificationRepository, org.mockito.Mockito.times(2)).save(any(CaseNotification.class));
```
trevorhinson, line 85: Static import on the times(2) maybe?

src/test/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationServiceTest.java
```diff
@@ -199,4 +202,41 @@ void testCheckNotificationStatusError() throws NotificationClientException {
 
         verify(notificationClient).getNotificationById(notificationId);
     }
+
+    @DisplayName("Should generate UUID for case ID when sending email")
+    @Test
+    void shouldGenerateUuidForCaseIdWhenSendingEmail() throws NotificationClientException {
+        // Given
```
trevorhinson, line 209: Not sure if there is a coding standard on this as I'd applied GWT on one of my tickets in the past and then they were removed in another.   The rest of the class does not have them though so perhaps remove for consistency?

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
```diff
@@ -55,6 +60,18 @@ public SendEmailResponse sendEmail(EmailNotificationRequest emailRequest) {
...
+                caseNotification.setProviderNotificationId(providerNotificationId);
+                caseNotification.setStatus(NotificationStatus.SENDING.toString());
+                caseNotification.setSubmittedAt(LocalDateTime.now());
+                notificationRepository.save(caseNotification);
+                log.info("Updated notification with provider ID: {}", providerNotificationId);
+            } catch (Exception e) {
+                log.error("Error updating notification with provider ID: {}", e.getMessage(), e);
+            }
+            
             log.debug("Email sent successfully. Reference ID: {}", referenceId);
```
trevorhinson, line 75: The log here can be misleading if a failure if possible in the new addition above.

src/integrationTest/java/uk/gov/hmcts/reform/pcs/notify/NotificationEndpointIT.java
```diff
@@ -106,7 +106,7 @@ void shouldSaveNotificationWhenEndpointGetsCalled() throws Exception {
         List<CaseNotification> notifications = notificationRepository.findAll();
         assertThat(notifications).isNotEmpty();
         assertThat(notifications.getFirst().getRecipient()).isEqualTo(request.getEmailAddress());
-        assertThat(notifications.getFirst().getStatus()).isEqualTo("pending-schedule");
+        assertThat(notifications.getFirst().getStatus()).isEqualTo("sending");
```
trevorhinson, line 109: Perhaps use the NotificationStatus enum here?

src/test/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationServiceTest.java
andrewdami11: I normally add it out of habit, and to keep me in track when writing tests. Happy to remove it, to keep it consistent though :-)

src/integrationTest/java/uk/gov/hmcts/reform/pcs/notify/NotificationEndpointIT.java
andrewdami11: Addressed [here]([refactor: replace string literals with NotificationStatus enums, in N…](https://github.com/hmcts/pcs-api/pull/186/commits/370abd88d67475b45a5a5aa041cddd141886cde1))

src/test/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationServiceTest.java
andrewdami11: Addressed [here](https://github.com/hmcts/pcs-api/pull/186/commits/d463f8d791e059e531a2d2f04b9526604de565af)

src/test/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationServiceTest.java
andrewdami11: Addressed [here](https://github.com/hmcts/pcs-api/pull/186/commits/2e8b23d35add8541dbc778c9972c0d4cf74647bb)

src/integrationTest/java/uk/gov/hmcts/reform/pcs/notify/NotificationEndpointIT.java
andrewdami11: Addressed [here](https://github.com/hmcts/pcs-api/pull/186/commits/18cd854afb20cfec6d4cbef55353b3d27d47a9eb)

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
andrewdami11: Good catch 💡
Addressed [here](https://github.com/hmcts/pcs-api/pull/186/commits/bec43b54a682c12fbb0c04f961d78c8a3c096720)

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
andrewdami11: Hi Scott, in the `NotificationStatus` class, `toString()` has been overridden to perform the same operation as `getValue()`, hence the confusion, I think. 
I have refactored the method used, for better clarity [here](https://github.com/hmcts/pcs-api/pull/186/commits/ad717bb242e61859479a6b179799ba5e514ff3d6)

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
andrewdami11: Addressed [here](https://github.com/hmcts/pcs-api/pull/186/commits/fc97b30aadca3ea1a1e714d821fd0452da092e8c)

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
```diff
@@ -56,6 +61,10 @@ public SendEmailResponse sendEmail(EmailNotificationRequest emailRequest) {
             );
 
             log.debug("Email sent successfully. Reference ID: {}", referenceId);
+
+            // Update notification with provider ID received from GOV.UK Notify
+            UUID providerNotificationId = UUID.fromString(sendEmailResponse.getNotificationId().toString());
+            updateNotificationStatus(caseNotification, NotificationStatus.SENDING, providerNotificationId);
```
adusumillipraveen, line 67: I think this should be NotificationStatus.SUBMITTED (a status on our side), (sending might come from gov notify response in get status calls)

andrewdami11: @adusumillipraveen where (or who) should I get confirmation of what exactly this should be?

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
```diff
@@ -127,4 +158,40 @@ CaseNotification createCaseNotification(String recipient, String type, UUID case
...
+
+    /**
+     * Updates the status of a notification and saves it to the database.
+     * 
+     * @param notification The notification to update
+     * @param status The new status to set
+     * @param providerNotificationId Optional provider notification ID to set (can be null)
+     * @return The updated notification (or null if an error occurred)
+     */
+    private CaseNotification updateNotificationStatus(
```
trevorhinson, line 170: I'd be tempted to have this package private and have a test around it.   There are a number of different exits from it.

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
```diff
@@ -127,4 +158,40 @@ CaseNotification createCaseNotification(String recipient, String type, UUID case
...
+            CaseNotification notification, 
+            NotificationStatus status, 
+            UUID providerNotificationId) {
+        
+        try {
+            notification.setStatus(status.getValue());
+            notification.setLastUpdatedAt(LocalDateTime.now());
+            
+            if (providerNotificationId != null) {
```
trevorhinson, line 179: why the if statement?

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
adusumillipraveen: I can confirm :)

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
scottstewart-sl: @andrewdami11 Actually I was wondering why the `CaseNotification` entity isn't using the enum type directly, (and writing the enum value directly to the DB)? It would be better if we could keep the value in its enum form as much as possible and then only convert to the string value where that needs to be shown to a user or sent to another system.

P.S. By enum value I mean the enum entry, e.g. `PENDING_SCHEDULE`. It's slightly confusing that the string field in the enum is also called `value` because it overlaps with the built in enum method `values()`.

src/test/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationServiceTest.java
```diff
@@ -199,4 +203,38 @@ void testCheckNotificationStatusError() throws NotificationClientException {
 
         verify(notificationClient).getNotificationById(notificationId);
     }
+
+    @DisplayName("Should generate UUID for case ID when sending email")
+    @Test
+    void shouldGenerateUuidForCaseIdWhenSendingEmail() throws NotificationClientException {
```
trevorhinson, line 209: Not sure on the need to specify the Exception type.  Just throw Exception no?

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
scottstewart-sl: Personally I would say that as long as it is easy get reasonable coverage of the paths with the unit tests coming in via the public methods then it is ok to keep it private.

And if it were to be sufficiently complicated to need calling directly for unit test scenarios then it would be better in a separate class and injected as a dependency. That way tests can also be added to ensure that it is being called with the correct params from this class.

trevorhinson, line 170: The CaseNotification can be returned as null from here so perhaps return an Optional<CaseNotification> or make this void?

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
```diff
@@ -84,10 +96,29 @@ public CompletableFuture<Notification> checkNotificationStatus(String notificati
                 TimeUnit.MILLISECONDS.sleep(statusCheckDelay);
                 
                 Notification notification = notificationClient.getNotificationById(notificationId);
-                log.info("Notification status check - ID: {}, Status: {}", 
-                    notificationId,
-                    notification.getStatus()
-                );
+                String status = notification.getStatus();
```
trevorhinson, line 99: Think this can be moved down to closer to where it is used.

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
andrewdami11: @adusumillipraveen Good enough for me 🙃

andrewdami11: @adusumillipraveen are you suggesting to replace the `NotificationStatus.SENDING` enum to `NotificationStatus.SUBMITTED` instead, throughout the codebase?

adusumillipraveen: No, only on this case when the notification has been successfully sent to Gov notify.

src/test/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationServiceTest.java
andrewdami11: `NotificationClientException` is more precise and better aligns with the actual exception that might occur.

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
andrewdami11: Okay, I will need to add the `SUBMITTED` enum value to the NotificationStatus class then, in that case

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
andrewdami11: Ah yes, good observation

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
```diff
@@ -56,6 +61,10 @@ public SendEmailResponse sendEmail(EmailNotificationRequest emailRequest) {
             );
 
             log.debug("Email sent successfully. Reference ID: {}", referenceId);
+
+            // Update notification with provider ID received from GOV.UK Notify
+            UUID providerNotificationId = UUID.fromString(sendEmailResponse.getNotificationId().toString());
```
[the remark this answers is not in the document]
andrewdami11: Ah yes, I left this incomplete. Addressed [here](https://github.com/hmcts/pcs-api/pull/186/commits/1985761676ef7b631a1c42e3b13b84a880a2192c)

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
andrewdami11: Addressed [here](https://github.com/hmcts/pcs-api/pull/186/commits/84861225b60b45ca3542c93062807f352aff2a6b)

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
andrewdami11: Oh yes, I now understand what is it that you meant.
Addressed [here](https://github.com/hmcts/pcs-api/pull/186/commits/ea9b3cede09ba088ca8ac5ba48ce4279db0f4254)

src/test/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationServiceTest.java
TrevorHinsonHMCTS: 👍   Ah fair enough argument.   I've said the same in the past and been shot down for it as this is not verifying specific exception behaviour.  The test is not being used elsewhere also.  Anyway I'm cool with this.

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
TrevorHinsonHMCTS: Does not completely address it.   The exception thrown will result in a null still on the last line of the method.

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
```diff
@@ -127,4 +159,40 @@ CaseNotification createCaseNotification(String recipient, String type, UUID case
...
         }
     }
+
+    /**
+     * Updates the status of a notification and saves it to the database.
+     * 
+     * @param notification The notification to update
+     * @param status The new status to set
+     * @param providerNotificationId Optional provider notification ID to set (can be null)
+     * @return The updated notification (or null if an error occurred)
```
trevorhinson, line 169: @return An Optional containing the updated notification, or an empty Optional if an error occurred

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
```diff
@@ -84,10 +97,29 @@ public CompletableFuture<Notification> checkNotificationStatus(String notificati
                 TimeUnit.MILLISECONDS.sleep(statusCheckDelay);
                 
                 Notification notification = notificationClient.getNotificationById(notificationId);
-                log.info("Notification status check - ID: {}, Status: {}", 
-                    notificationId,
-                    notification.getStatus()
-                );
+                
+                // Find and update notification status in database
+                try {
```
trevorhinson, line 102: There is a question as to if this try block should be within its' own method now.

andrewdami11: That's fair. 
I've extracted it in [this commit](https://github.com/hmcts/pcs-api/pull/186/commits/9a681ec06083fc95142b1093619f7c407e09ba8d), and added relevant tests in [this other commit](https://github.com/hmcts/pcs-api/pull/186/commits/f47fd59dfa7d3eb430025bd9502c351a5e383c8a)

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
```diff
@@ -84,10 +97,29 @@ public CompletableFuture<Notification> checkNotificationStatus(String notificati
...
+                    CaseNotification caseNotification = notificationRepository
+                        .findByProviderNotificationId(UUID.fromString(notificationId))
+                        .orElse(null);
+                    
+                    if (caseNotification != null) {
+                        String status = notification.getStatus();
+                        try {
```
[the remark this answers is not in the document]
andrewdami11: Similarly, I've extracted it in [this commit](https://github.com/hmcts/pcs-api/pull/186/commits/9a681ec06083fc95142b1093619f7c407e09ba8d), and added relevant tests in [this other commit](https://github.com/hmcts/pcs-api/pull/186/commits/f47fd59dfa7d3eb430025bd9502c351a5e383c8a)

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
andrewdami11: Addressed [here](https://github.com/hmcts/pcs-api/pull/186/commits/22288cf6c91508d68b8bea1f35c5564fccfe3327)

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
andrewdami11: @TrevorHinsonHMCTS what would you suggest?

scottstewart-sl (dismissed):

adusumillipraveen (approved):

adusumillipraveen (dismissed):

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
```diff
@@ -127,4 +154,115 @@ CaseNotification createCaseNotification(String recipient, String type, UUID case
...
+            }
+            
+            CaseNotification saved = notificationRepository.save(notification);
+            log.info("Updated notification status to {} for notification ID: {}", 
+                    status, notification.getNotificationId());
+            return Optional.of(saved);
+        } catch (Exception e) {
+            log.error("Error updating notification status to {}: {}", 
+                    status, e.getMessage(), e);
+            return null;
```
trevorhinson, line 265: Here is the difference.

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
TrevorHinsonHMCTS: Sorry was mentioned in the Slack.   I think this should return Optional.empty() instead of the null at the end of the method

MoLala1 (approved):

scottstewart-sl (approved):

PR #192, opened 2025-05-08
Hdpi 608 update delivery status to db

jaina-sol:
### Jira link

See [HDPI-608](https://tools.hmcts.net/jira/browse/HDPI-608)

#### This PR implementation tracks notification status, updating the database with the current status from Gov Notify.
It captures the notification status (delivered, failed, pending, etc) in the database and implements a status checking mechanism that updates the records when the status changes.

### Checklist

- [x] Does this PR introduce a breaking change


src/main/java/uk/gov/hmcts/reform/pcs/notify/model/NotificationStatus.java
```diff
@@ -0,0 +1,36 @@
...
+    CREATED("created"),
+    SENDING("sending"),
+    DELIVERED("delivered"),
+    PERMANENT_FAILURE("permanent-failure"),
+    TEMPORARY_FAILURE("temporary-failure"),
+    TECHNICAL_FAILURE("technical-failure"),
+    SCHEDULE("scheduled"),
+    PENDING_SCHEDULE("pending-schedule");
+
+    private final String value;
```
scottstewart-sl, line 13: I've seen these changes in another PR somewhere. The use of the name `value` for this field is a bit ambiguous because really the "values" of an enum are the enum entries, e.g. `DELIVERED` as type NotificationStatus. (Hence the enum methods `NotificationStatus.values()` and `valueOf()`).

What do "delivered", "technical-failure" etc represent? Is that the representation of each state in the API response from the Notification Service? If so, then perhaps the field could be called `notificationServiceValue` or `govNotifyValue` or something like that. Then the getter would be a bit more meaningful and you could use it in place of `toString()` in the code.

src/main/resources/db/migration/V002__case_notification.sql
scottstewart-sl, line 1: Just for info - there are 2 PRs going through at the moment that add a V002 migration, so whichever one doesn't get to the end first will need to update to V003 😄

src/main/java/uk/gov/hmcts/reform/pcs/notify/domain/CaseNotification.java
```diff
@@ -0,0 +1,69 @@
...
+@EqualsAndHashCode(of = "notificationId")
+public class CaseNotification {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.UUID)
+    @Column(name = "notification_id", nullable = false)
+    private UUID notificationId;
+
+    @Column(name = "case_id", nullable = false)
+    private UUID caseId;
```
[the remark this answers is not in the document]
scottstewart-sl: Btw, I'm happy to have a discussion in the team about this. I can see that using the CCD case reference as our primary case identifier would be easier in some cases. I've just learnt from past experience that it can be simpler to have a primary key that has not come from what is effectively an external system, particularly when trying to repair data.

src/main/resources/db/migration/V002__case_notification.sql
scottstewart-sl: The other PR that added a migration has been merged, (and had 2 migrations actually), so this will need to be renamed to V004

PR #195, opened 2025-05-08
HDPI-661: Remove spaces in postcode mappings

scottstewart-sl:
### Jira link

See [HDPI-661](https://tools.hmcts.net/jira/browse/HDPI-661)

### Change description

Remove whitespace in postcode mappings, including spaces within the postcode, at the start and at the end. For now, the integration tests that calls the postcode court endpoint has been modified to use postcodes with no space - this can be reverted once HDPI-709 is done.

The main change here is the addition of a DB trigger. Ideally the test for this would be a DB unit test, but given that hopefully we won't have much logic in the DB, (e.g. in stored procedures), it is not worth the effort of setting up such a framework for now and I have used a Spring Boot test instead to check the DB trigger behaviour.

I've also updated the integration tests to provide their own test data and not rely on the post code test data that is intended for the preview environment.

### Testing done

Tests updated and run locally.

### Checklist


src/integrationTest/resources/data.sql
```diff
@@ -0,0 +1,7 @@
+INSERT INTO postcode_court_mapping (postcode, epimid, legislative_country, effective_from, effective_to, audit)
+VALUES ('W3 7RX', 20262, 'England', '2025-01-01', '2035-04-01', '{}'::jsonb),
+       ('W3 6RS', 36791, 'England', '2025-02-01', NULL, '{}'::jsonb),
+       ('M13 9PL', 144641, 'England', '2025-04-01', '2035-04-01', '{}'::jsonb),
+       ('TST1 9BC', 1000, 'England', '2025-01-01', NULL, '{}'::jsonb),
+       ('tst2 9BC', 1000, 'England', '2025-01-01', '2025-12-31', '{}'::jsonb),
```
isha2040, line 6: Can we not use the same sample data here for integration tests ?

isha2040: ok, that makes sense. thank you

isha2040 (approved):

src/main/resources/db/migration/R__postcode_court_triggers.sql
andrewdami11, line 1: Is it worth adding NULL handling to your trigger function, before the standardization logic? For example:

IF NEW.postcode IS NULL THEN
  RETURN NEW;
END IF;

This would avoid potential issues if any NULL postcodes make their way into the system.

src/main/resources/db/migration/R__postcode_court_triggers.sql
```diff
@@ -0,0 +1,23 @@
...
+$func$;
+
+/*
+  CREATE OR REPLACE TRIGGER was only added in Postgres 14 and at the time of
+  writing the Cftlib Postgres used in local development is still on
+  version 12, (even though Azure DBs for the pcs-api are on 16).
+
+  When Cftlib is using Postgres 14+ then the DROP/CREATE can be
+  changed to a CREATE OR REPLACE
+*/
```
andrewdami11, line 18: While these comments about Postgres version constraints are helpful, is it worth adding a header comment explaining the purpose of this trigger and implementation date?

src/integrationTest/java/uk/gov/hmcts/reform/pcs/postcodecourt/PostCodeDatabaseIT.java
andrewdami11, line 1: I understand the tradeoff between setting up a DB unit test framework versus using Spring Boot tests, that you've mentioned in the PR description. Is it worth adding a comment here, explaining this testing decision for future reference?

src/main/resources/db/migration/R__postcode_court_triggers.sql
scottstewart-sl: The postcode column has a not null constraint on it, so it shouldn't even be null when this trigger is called. (I've tested this locally as well).

src/integrationTest/java/uk/gov/hmcts/reform/pcs/postcodecourt/PostCodeDatabaseIT.java
scottstewart-sl: I've added a comment for this now

src/main/resources/db/migration/R__postcode_court_triggers.sql
scottstewart-sl: Actually Jake has kindly updated the CftLib to use Postgres 16 so this comment is no longer relevant. I've removed it and added a general header comment about the purpose of the trigger.

Personally I'm not a fan of adding a creation or update date to the header comments because these are often not maintained and the same information can generally be found from the Git history.

JakeCG (approved):

PR #224, opened 2025-05-27
HDPI 662 - Implement shortened exact string match logic for postcode lookup

libanAbdirahman1:
https://tools.hmcts.net/jira/browse/HDPI-662

### Change description

This change introduces partial postcode lookup logic for the postcode court mapping table. If no exact match is found for a full postcode, the last character of the postcode will be trimmed (up to three times) and lookup is reattempted to find an exact string match.

### Testing done
- Manual testing performed using various full and partial postcodes.
- Unit and Integration tests added/updated to cover the new partial lookup logic.

### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
  # renovate: datasource=github-releases depName=microsoft/ApplicationInsights-Java
-ARG APP_INSIGHTS_AGENT_VERSION=3.7.2
+ARG APP_INSIGHTS_AGENT_VERSION=3.7.3
 FROM hmctspublic.azurecr.io/base/java:21-distroless
 
 COPY lib/applicationinsights.json /opt/app/
--- a/build.gradle
+++ b/build.gradle
@@ -281,7 +281,7 @@ repositories {
 ext {
   log4JVersion = "2.24.3"
   logbackVersion = "1.5.18"
-  testcontainersVersion = "1.21.0"
+  testcontainersVersion = "1.21.1"
   springFrameworkBootVersion = "3.4.5"
   lombokVersion = "1.18.38"
   springSecurityVersion = "6.5.0"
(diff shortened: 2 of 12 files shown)


src/test/java/uk/gov/hmcts/reform/pcs/postcodecourt/exception/PostCodeExceptionHandleTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+    @Test
+    void shouldReturnInternalServerErrorForNotificationException() {
+        PostCodeNotFoundException exception = mock(PostCodeNotFoundException.class);
+
+        when(exception.getMessage()).thenReturn("No court mapping found for postcode");
+        ResponseEntity<PostCodeExceptionHandler.Error> response = handler.handlePostCodeNotFoundException(exception);
+
+        assertEquals(HttpStatus.NOT_FOUND, response.getStatusCode());
+        assertNotNull(response.getBody());
```
scottstewart-sl, line 25: It's preferable to use AssertJ for assertions.

src/main/java/uk/gov/hmcts/reform/pcs/postcodecourt/service/PostCodeCourtService.java
```diff
@@ -49,9 +52,29 @@ private List<PostCodeCourtEntity> getPostcodeCourtMappings(String postcode) {
...
+        for (int i = 0; i < maxTrim && trimmedPostcode.length() > 1; i++) {
+            trimmedPostcode = trimmedPostcode.substring(0, trimmedPostcode.length() - 1);
+            log.info("Attempting partial match lookup for trimmed postcode: {}", trimmedPostcode);
+            List<PostCodeCourtEntity> result = postCodeCourtRepository.findByIdPostCode(trimmedPostcode);
+            if (!result.isEmpty()) {
+                log.info("Found partial match of {} for postcode {}", trimmedPostcode, postcode);
+                return result;
+            }
+        }
+        throw new PostCodeNotFoundException("No court mapping found for postcode " + postcode);
```
scottstewart-sl, line 77: What do you think about moving this `throw` up into the calling method? Arguably this `tryPartialPostCodeLookup` method isn't concerned with what should happen if there are no matches - it's the `getCountyCourtsByPostCode` method above that decides than having no results should be an error.

src/main/java/uk/gov/hmcts/reform/pcs/postcodecourt/service/PostCodeCourtService.java
```diff
@@ -49,9 +52,29 @@ private List<PostCodeCourtEntity> getPostcodeCourtMappings(String postcode) {
...
-        return postCodeCourtRepository.findByIdPostCode(postcode);
+        List<PostCodeCourtEntity> courts = postCodeCourtRepository.findByIdPostCode(postcode);
+        if (courts.isEmpty()) {
+            log.warn("Full postcode match not found for {}, Attempting partial match", postcode);
+            courts = tryPartialPostCodeLookup(postcode);
+        }
+        return courts;
+    }
+
+    private List<PostCodeCourtEntity> tryPartialPostCodeLookup(String postcode) {
```
scottstewart-sl, line 64: Is there a clever way these look ups could be done with a single DB query? Otherwise, there could be up to 4 DB calls just to check one postcode.

scottstewart-sl:
I know the ticket says that if there is no match for the postcode then the api should respond with an error, but I think we should double check that requirement. For an endpoint that returns a list, an empty list would also be a good indication of no matching results and in my opinion is more consistent with the expected behaviour of a search type endpoint.

Also before these changes, the endpoint returns an empty list for an unknown postcode, so this ticket is changing that behaviour.

scottstewart-sl (approved):

PR #235, opened 2025-06-02
HDPI-753 Enhancements to courts endpoint

libanAbdirahman1:
[HDPI-753](https://tools.hmcts.net/jira/browse/HDPI-753)

### Change description

Added validation for IDAM token and postcode param on /courts endpoints.
- Returns 401 when IDAM token is invalid, expired, or missing.
- Returns 400 when postcode is null, empty, or not provided.
- Ensures valid tokens and postcodes return 200 when matched.

### Testing done

- Unit and Integration tests were added.
- Manual testing done for IDAM tokens and postcodes via postman.

### Checklist

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/pcs/config/IdamAuthenticationFilter.java
+++ b/src/main/java/uk/gov/hmcts/reform/pcs/config/IdamAuthenticationFilter.java
@@ -1,0 +1,54 @@
+package uk.gov.hmcts.reform.pcs.config;
+
+import jakarta.servlet.FilterChain;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.stereotype.Component;
+import org.springframework.web.filter.OncePerRequestFilter;
+import uk.gov.hmcts.reform.pcs.exception.InvalidAuthTokenException;
+import uk.gov.hmcts.reform.pcs.idam.IdamService;
+import uk.gov.hmcts.reform.pcs.idam.User;
+
+import java.io.IOException;
+import java.util.Collections;
+
+@Slf4j
+@Component
+public class IdamAuthenticationFilter extends OncePerRequestFilter {
+
+    @Autowired
+    private  IdamService idamService;
+
+    @Override
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response,
+                                    FilterChain filterChain) throws ServletException, IOException {
+        String path = request.getRequestURI();
+
+        if (!path.startsWith("/courts")) {
+            filterChain.doFilter(request, response);
+            return;
+        }
+        String authToken = request.getHeader("Authorization");
+
+        try {
+            User user = idamService.validateAuthToken(authToken);
+            Authentication authentication =
+                new UsernamePasswordAuthenticationToken(user, null, Collections.emptyList());
+            SecurityContextHolder.getContext().setAuthentication(authentication);
+            filterChain.doFilter(request, response);
+        } catch (InvalidAuthTokenException ex) {
+            log.error("Authentication failed: {}", ex.getMessage(), ex);
+            response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+            response.setContentType("application/json");
+            String json = "{\"error\":\"" + ex.getMessage() + "\"}";
+            response.getWriter().write(json);
+        }
+    }
+
+}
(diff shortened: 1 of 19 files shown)


src/main/java/uk/gov/hmcts/reform/pcs/config/IdamAuthenticationFilter.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response,
+                                    FilterChain filterChain) throws ServletException, IOException {
+        String path = request.getRequestURI();
+
+        if (!path.startsWith("/courts")) {
+            filterChain.doFilter(request, response);
+            return;
+        }
+        String authToken = request.getHeader("Authorization");
```
scottstewart-sl, line 37: Minor - you could use `HttpHeaders.AUTHORIZATION` to get this header name

src/main/java/uk/gov/hmcts/reform/pcs/idam/IdamService.java
```diff
@@ -0,0 +1,55 @@
...
+import uk.gov.hmcts.reform.pcs.exception.InvalidAuthTokenException;
+
+import static com.azure.spring.cloud.autoconfigure.implementation.aad.security.constants.Constants.BEARER_PREFIX;
+
+@Service
+@Slf4j
+public class IdamService {
+
+    @Autowired
+    private IdamClient idamClient;
```
scottstewart-sl, line 18: Constructor injection is preferable to field injection.

src/main/resources/application.yaml
```diff
@@ -82,6 +82,13 @@ idam:
     microservice: ${S2S_SERVICE_NAME:pcs_api}
   s2s-authorised:
     services: ${S2S_NAMES_WHITELIST:pcs_api,pcs_frontend,ccd_data}
+  systemupdate:
+    username: ${IDAM_SYSTEM_USERNAME:dummysystemupdate@test.com}
+    password: ${IDAM_SYSTEM_PASSWORD:dummy}
+  client:
+    id: 'pcs'
```
scottstewart-sl, line 89: I think the IDAM client ID is `pcs-api`. (I discovered this after trying to track down why my own PR wasn't working in Preview 😄  )

src/test/java/uk/gov/hmcts/reform/pcs/idam/IdamServiceTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Mock
+    private IdamClient idamClient;
+
+    @InjectMocks
+    private IdamService underTest;
+
+    @DisplayName("Should throw InvalidAuthTokenException when token is null")
+    @Test
+    void shouldThrowInvalidAuthTokenExceptionWhenAuthTokenIsNull() {
```
scottstewart-sl, line 34: Minor, you could use a parameterised test here to check null, empty and blank auth tokens in a single test method. Up to you.

src/main/java/uk/gov/hmcts/reform/pcs/config/IdamAuthenticationFilter.java
```diff
@@ -0,0 +1,54 @@
...
+            Authentication authentication =
+                new UsernamePasswordAuthenticationToken(user, null, Collections.emptyList());
+            SecurityContextHolder.getContext().setAuthentication(authentication);
+            filterChain.doFilter(request, response);
+        } catch (InvalidAuthTokenException ex) {
+            log.error("Authorization failed: {}", ex.getMessage(), ex);
+            response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+            response.setContentType("application/json");
+            String json = "{\"error\":\"" + ex.getMessage() + "\"}";
+            response.getWriter().write(json);
```
scottstewart-sl, line 50: In theory, this filter doesn't know what the content type of the response it, so it is probably better to just return the status code with an empty body, (like `ServiceAuthFilter`). Also it would be more secure to not provide any information about why the token was invalid, and instead just have it in the logs.

MoLala1 (approved):

src/main/java/uk/gov/hmcts/reform/pcs/config/IdamAuthenticationFilter.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Autowired
+    private  IdamService idamService;
+
+    @Override
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response,
+                                    FilterChain filterChain) throws ServletException, IOException {
+        String path = request.getRequestURI();
+
+        if (!path.startsWith("/courts")) {
```
scottstewart-sl, line 33: Is there way to just configure which paths this filter applies to in the SecurityConfiguration class, rather than list the endpoints here?

src/main/java/uk/gov/hmcts/reform/pcs/config/IdamAuthenticationFilter.java
scottstewart-sl: Maybe something like this would help: https://stackoverflow.com/a/74389699

src/main/java/uk/gov/hmcts/reform/pcs/config/IdamAuthenticationFilter.java
scottstewart-sl: Maybe something like this would help: https://stackoverflow.com/a/74389699

src/main/java/uk/gov/hmcts/reform/pcs/config/IdamAuthenticationFilter.java
libanAbdirahman1: I agree, i've removed the unneccessary responses now.

scottstewart-sl (approved):

PR #301, opened 2025-06-23
HDPI-660: Add script to validate and convert postcode/ePIMS mappings

scottstewart-sl:
### Jira link

See [HDPI-660](https://tools.hmcts.net/jira/browse/HDPI-660)

### Change description

It is assumed for now that the mappings from postcodes and partial postcodes to court ePIMS IDs will be provided in a CSV format. This change adds a sample CSV format and a simple script to validate the mappings before generating the necessary SQL inserts to replace the data in the `postcode_court_mapping` table.

This is what the generated SQL looks like for the sample CSV data that is included with this PR:

```
Postcode,e-PIMS ID,Legislative Country,Effective From,Effective To
W3 7RX,20262,England,30/01/2024,25/10/2035
W3 7RX,20263,England,26/10/2035,25/10/2039
W3 6RS,36791,England,15/01/2000,
W5 7,144641,England,15/01/2000,null
```

```
BEGIN;
TRUNCATE postcode_court_mapping;
INSERT INTO postcode_court_mapping (postcode, epimid, legislative_country, effective_from, effective_to, audit)
VALUES
   ('W3 7RX', 20262, 'England', '2024-01-30', '2035-10-25', '{"generated": "2025-06-23T10:02:39.968Z" }'),
   ('W3 7RX', 20263, 'England', '2035-10-26', '2039-10-25', '{"generated": "2025-06-23T10:02:39.968Z" }'),
   ('W3 6RS', 36791, 'England', '2000-01-15', NULL, '{"generated": "2025-06-23T10:02:39.968Z" }'),
   ('W5 7', 144641, 'England', '2000-01-15', NULL, '{"generated": "2025-06-23T10:02:39.968Z" }');
COMMIT;
```

### Testing done

Manual testing

  * [ ] Yes
  * [x] No

### Checklist


MoLala1 (approved):

tools/postcode-to-epims/src/csv-parser.js
```diff
@@ -0,0 +1,67 @@
...
+    return null;
+  }
+
+  return DateTime.fromFormat(sourceDateString, 'dd/MM/yyyy');
+}
+
+export const readCsvData = (csvSourceFile) => {
+  let csvData;
+  try {
+    csvData = fs.readFileSync(csvSourceFile);
```
paddy-hmcts, line 16: might have been better to use streams and pipe those directly into the csv parse lib (granted the csv files are only small though so just an improvement suggestion) e.g. https://www.digitalocean.com/community/tutorials/how-to-read-and-write-csv-files-in-node-js-using-node-csv

tools/postcode-to-epims/src/sql-generator.js
```diff
@@ -0,0 +1,26 @@
+import { DateTime } from 'luxon';
+import fs from 'fs';
+import nunjucks from 'nunjucks'
```
paddy-hmcts, line 3: I really wouldn't have used nunjucks here to purely generate essentially what are prepared statements, maybe have a look at a db library to do this heavy lifting for you

paddy-hmcts (changes requested):
On the whole seems 'fine', but I do wonder if there was perhaps a java way of doing this (than having to introduce another tool such as node). I would definitely remove the nunjucks in favour of a db library (the stream stuff is optional but it would reduce memory consumption and can handle larger files if need be)

scottstewart-sl:
> On the whole seems 'fine', but I do wonder if there was perhaps a java way of doing this (than having to introduce another tool such as node).

@paddy-hmcts The requirements on how this data will be updated on an ongoing basis haven't been finalised yet so I just wanted to create a simple script for the one off initial seeding of the postcode data. If we were to be receiving regular CSV update files then I agree that a proper endpoint in the Java app to upload and validate the data would have made more sense. For now, it is just a tool and won't be part of the released code.

I could have written it in Java but I think doing some scripting like this is easier in Node and I don't see an issue with having a combination of languages in the repo. The e2e tests in this repo are also using Node to run the Playwright tests.

paddy-hmcts (approved):

tools/postcode-to-epims/src/sql-generator.js
scottstewart-sl: I've reverted this to using template literals now and removed the nunjucks dependency as suggested.

PR #324, opened 2025-07-01
Hdpi 879 create a document via doc assembly unhappy paths

MoLala1:
HDPI-879

This ticket should be merged into master ONLY after [HDPI-865](https://tools.hmcts.net/jira/browse/HDPI-865).
A PR can already be opened however, i am making sure it points to [HDPI-865](https://tools.hmcts.net/jira/browse/HDPI-865) (not master), so that only the relevant 4 files that have changed are displayed.


libanAbdirahman1 (approved):

src/test/java/uk/gov/hmcts/reform/pcs/document/service/DocAssemblyServiceTest.java
```diff
@@ -388,4 +390,254 @@ void shouldHandleValidRenditionOutputLocation() throws Exception {
...
+            eq(SERVICE_AUTH_TOKEN), 
+            any(DocAssemblyRequest.class)
+        )).thenThrow(feignException);
+
+        // When & Then
+        assertThatThrownBy(() -> docAssemblyService.generateDocument(request))
+            .isInstanceOf(DocAssemblyException.class)
+            .hasMessage("Bad request to Doc Assembly service: Bad request error")
+            .hasCause(feignException);
+    }
```
scottstewart-sl, line 417: There's quite a lot of duplication in these tests, which could lead to more of a maintenance overhead. From what I can see the sub-type of `FeignException` that is thrown doesn't actually matter - it is the response status code that is being used to control the handling of the exception. Also the `FeignException` message doesn't control anything either, so that could be any test value.

How about creating a single parameterised test that takes the response status and expected first part of the `DocAssemblyException` message and then covers all these cases?

JakeCG (approved):

PR #495, opened 2025-08-21
HDPI-1506: rent details

paddy-hmcts:
### Jira link

https://tools.hmcts.net/jira/browse/HDPI-1506


MoLala1 (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/createpossessionclaim/RentDetails.java
```diff
@@ -2,23 +2,32 @@
...
-                    Under Development
-                  """);
+                        ---
+                        <section tabindex="0">
+                            <p class="govuk-body">
+                                Please provide details about the current rental agreement.
+                            </p>
+                        </section>
+                        """)
+                .mandatory(PCSCase::getRentAmount)
```
AlexBuzea, line 28: Should this be mandatory in all cases except if they select Other?
I think you can add a conditional like this so if they select Other that field will no longer be mandatory like this:
`.mandatory(PCSCase::getRentAmount, "rentPaymentFrequency!=\"OTHER\"")`

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/createpossessionclaim/RentDetails.java
```diff
@@ -2,23 +2,32 @@
...
 public class RentDetails implements CcdPageConfiguration {
 
     @Override
     public void addTo(PageBuilder pageBuilder) {
         pageBuilder
                 .page("rentDetails")
-                .pageLabel("Rent details (placeholder)")
-                .label("rentDetails-info",
+                .pageLabel("Rent details")
+                .label("caseReference", "Case number:")
```
AlexBuzea, line 18: I'd probably remove this line for now like on the other pages. The case number is currently generated after the case is submitted

paddy-hmcts: yeah the figma design suggested we show the caseId, do we not generate a case first in draft mode and store it against the journey then? or is it all data submitted at the end?

PR #518, opened 2025-09-01
HDPI-1429: Split creation event into two and save unsubmitted answers

scottstewart-sl:
### Jira link

See [HDPI-1429](https://tools.hmcts.net/jira/browse/HDPI-1429)

### Change description

The claim creation journey has been split into two events, just after the postcode eligibility check. That first event now creates a CCD case reference and persists the property address and legislative country in the DB. The second event gathers the remainder of the claim details, and persists the un-submitted answers as JSON in a separate schema and table, so that the user can come back to them in a later session. When the user finally submits the answers from the second event, they are also persisted in the DB and removed from the draft table.

This is the screen that the user sees at the end of the first (creation) event:
first-time

If the user provides some answers for the second event but does not submit them, when they return they will see this screen:

resume

Upon resuming, the user is given the choice of keeping their previous answers or starting from fresh:

keep-answers-choice

**Note** I haven't yet changed the packages of the Page classes that were moved to the new event because this would have led to a lot of affected files in this PR, which is already quite large. I'll do that refactoring as part of a subsequent ticket.

### Testing done

Unit tests added for refactored and new code

  * [ ] Yes
  * [x] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/createpossessionclaim/ResumeClaim.java
```diff
@@ -0,0 +1,74 @@
...
+        long caseReference = details.getId();
+        PCSCase caseData = details.getData();
+
+        log.debug("Resuming claim - keep existing answers = " + caseData.getResumeClaimKeepAnswers());
+
+        if (caseData.getResumeClaimKeepAnswers() == YesOrNo.YES) {
+            unsubmittedCaseDataService.getUnsubmittedCaseData(caseReference)
+                .ifPresentOrElse(
+                    unsubmittedCaseData -> modelMapper.map(unsubmittedCaseData, caseData),
+                    () -> log.warn("No unsubmitted case data found for case " + caseReference)
```
AlexBuzea, line 65: It it worth throwing an error here instead of a log? If the draft data doesn't exist it just proceeds and they won't know from what I'm seeing here?

scottstewart-sl: It's a good point. I had deliberately decided to let the user carry on because they've said Yes to keeping the existing answers, but there are no existing answers so there's nothing to keep.

In practice I think the main way for this to occur would be if someone/something else completed the claim event concurrently between this user being shown the screen to decide whether to keep their answers and them making and submitting their choice.

My main objective was to avoid having an error that prevented the user from doing anything, but on reflection I could throw an error here and the user could just try again and choose "No" to keeping the answers, in order to be able to proceed.

I'll update the code now

MoLala1 (approved):

PR #647, opened 2025-09-25
HDPI-1932: Add additional reasons for possession page

scottstewart-sl:
### Jira link

See [HDPI-1932](https://tools.hmcts.net/jira/browse/HDPI-1932)

### Change description

Add additional reasons for possession page. I'm trying out using a Complex type to group the related fields as a potential step to reducing the number of top level fields in the `PCSCase` class.

image

### Testing done

Unit tests extended, full build run locally

  * [ ] Yes
  * [x] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/createpossessionclaim/ClaimingCosts.java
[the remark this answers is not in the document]
scottstewart-sl: The package name helps to differentiate between the page config classes and the domain folder. This particular class is just a placeholder, but I think the convention for these page classes is to name the class more or less after the page title, which makes it easier to associate the class with the pages in ExUI, so I don't think we should change the names of these classes.

I can see that it gets a bit more tricky when the domain class and the page class have an identical name and the page class references the domain class, because then you need to fully qualify the classname, but I don't think that is an issue here.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/ClaimEntity.java
```diff
@@ -53,10 +53,12 @@ public class ClaimEntity {
     @OneToMany(fetch = LAZY, cascade = ALL, mappedBy = "claim")
     @Builder.Default
     @JsonManagedReference
-    private Set<ClaimGroundEntity> claimGroundEntities = new HashSet<>();
+    private Set<ClaimGroundEntity> claimGrounds = new HashSet<>();
 
     private String summary;
 
+    private String additionalReasons;
```
tvr-hmcts, line 60: Does this need to be applied in the CCDCaseRepository also via the getSubmittedCase method?

tvr-hmcts: Unless it is that once a Claim is made then the case is not reopen through this route?

scottstewart-sl: This event wouldn't be invoked again once the claim has been made. There are likely to be other case events that modified some part of the case data, and if those needed to use the additional reasons, (e.g. to allow editing), then we could add the code to CCDCaseRepository at that stage.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/createpossessionclaim/ClaimingCosts.java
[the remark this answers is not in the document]
scottstewart-sl: This might be something that will improve as you gain more familiarity with the code. Also it should be fairly obvious whether you are looking at a domain class, (which will just contain annotated fields), and a page configuration, (which will extend `CcdPageConfiguration` and have an overridden `addTo` method).

guygrewal77 (approved):
General refactoring changes to be addressed separately.

tvr-hmcts (approved):
Not sure why this comment is not visible?

"They may be simplistic in nature but should there be coverage of this class?"
Reference the ShowConditions class

Other than that all looks good.

PR #655, opened 2025-09-26
HDPI-1776 Defendant-circumstances

libanAbdirahman1:
### Jira link
https://tools.hmcts.net/jira/browse/HDPI-1776

### Change description

Adding initial screens to capture defendant circumstances

### Testing done
Unit test added
Manual testing done  
 
image

  * [ ] Yes
  * [x] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/PcsCaseEntity.java
```diff
@@ -88,6 +89,10 @@ public class PcsCaseEntity {
     @JdbcTypeCode(SqlTypes.JSON)
     private List<Defendant> defendants;
 
+    @Column(name = "defendant_circumstances")
+    @JdbcTypeCode(SqlTypes.JSON)
+    private DefendantCircumstancesInfo defendantCircumstancesInfo;
```
scottstewart-sl, line 94: According to the draft data model, defendant circumstances is a column on the `claim` table so you can just add it there instead of creating a JSON column for it in `pcs_case`.

I also don't think we need to store a flag to say whether info was provided or not - that can be determined by checking the DB column for null.

Btw, there was a discussion yesterday on whether we should use these Yes/No answers to validate the data coming from CCD, (i.e. if is "Yes" then validate that the string is not empty). I'll tag you in that conversation so you can read it.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/resumepossessionclaim/DefendantsDetails.java
```diff
@@ -70,6 +73,10 @@ private AboutToStartOrSubmitResponse<PCSCase, State> midEvent(CaseDetails<PCSCas
             }
         }
 
+        // TODO: Update this once multiple defendant support is implemented.
+        //  Set the text dynamically for one/multiple defendants.
+        caseData.setDynamicDefendantText("defendants'");
```
scottstewart-sl, line 78: Very minor point, but if we only support a single defendant at the moment then I would expect this to be "defendant's"

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/PCSCase.java
```diff
@@ -566,4 +566,24 @@ public class PCSCase {
...
+        typeOverride = TextArea,
+        max = 950,
+        access = {CitizenAccess.class}
+    )
+    private String defendantCircumstancesInfo;
+
+    @CCD(
+        access = {CitizenAccess.class}
+    )
+    private String dynamicDefendantText;
```
scottstewart-sl, line 587: For #647 I'm trying encapsulating the related fields into a complex type to keep `PcsCase` a bit cleaner and also to give some scope/context to these transitive fields like `dynamicDefendantText` that are only used on one page.

https://github.com/hmcts/pcs-api/blob/HDPI-1932_additional_reasons_for_possession/src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/AdditionalReasons.java

What do you think about trying the same thing here to wrap up these three fields?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/resumepossessionclaim/DefendantsDetails.java
libanAbdirahman1: I spoke to Chris about this earlier and he told me this is the default for the time being as per the designs. Possibly will be looked at again when the multiple defendants implementation is added and the dynamic text can be generated as per the number of defendants.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/PCSCase.java
libanAbdirahman1: I think thats a valid approach to make the pcs case abit cleaner. I have included a complex type for the defendant circumstances that wraps two of the fields. However the `dynamicDisplayedDefendantText` is still on the PcsCase as it will be called on an earlier page to populate it.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/DefendantCircumstancesInfo.java
```diff
@@ -0,0 +1,19 @@
...
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class DefendantCircumstancesInfo {
+
+    private Boolean infoProvided;
```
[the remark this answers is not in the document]
libanAbdirahman1: thats a valid point but this class has been removed now as the implementation has changed to add the defendant circumstances to the claim table.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/PCSCase.java
```diff
@@ -566,4 +566,11 @@ public class PCSCase {
     @CCD(access = {CitizenAccess.class})
     private NoRentArrearsReasonForGrounds noRentArrearsReasonForGrounds;
 
+    @JsonUnwrapped
+    @CCD(access = {CitizenAccess.class})
+    private DefendantCircumstances defendantCircumstances;
+
+    @CCD(access = {CitizenAccess.class})
+    private String dynamicDisplayedDefendantText;
```
scottstewart-sl, line 574: I was wondering if you could encapsulate this in the DefendantCircumstances complex type as well. At this level, it wouldn't be clear to someone reading it out of context whether it contains the dynamic name of the defendant or just the possessive form of the Defendant term.

There is a similar change in #606 for the Claimant Circumstances, but that one is showing the actual claimant name in possessive form: https://github.com/hmcts/pcs-api/blob/HDPI-1774-Claimant-Circumstances/src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/ClaimantCircumstances.java#L27

In fact, perhaps it would be clearer if it was called `defendantTermPossessive` or something like that. And if you could get it into the complex type as well then that would be more consistent with the ClaimantCircumstances

scottstewart-sl (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/ResumePossessionClaim.java
```diff
@@ -220,8 +223,12 @@ private void submit(EventPayload<PCSCase, State> eventPayload) {
 
         List<ClaimGroundEntity> claimGroundEntities =
             claimGroundService.getGroundsWithReason(pcsCase);
+        DefendantCircumstances defendantCircumstances = pcsCase.getDefendantCircumstances();
 
         claimEntity.addClaimGroundEntities(claimGroundEntities);
+        claimEntity.setDefendantCircumstances(
```
tvr-hmcts, line 229: Should there be some coverage of this within the ResumePossessionClaimTest?

tvr-hmcts (approved):

PR #1618, opened 2026-04-02
Hdpi 5499 service call back implementation

tvr-hmcts:
HDPI-5499 - [### Jira link](https://tools.hmcts.net/jira/browse/HDPI-5499)

### Change description

This PR implements the payment callback flow for the Fees & Pay service, enabling the application to receive and persist payment status updates from the payments service.

Adds a Flyway migration (V075__fee_payment.sql) creating a fee_payment table to persist payment records, linked to claim and party via foreign keys.

FeePaymentEntity — JPA entity mapped to the new fee_payment table, storing payment amount, status, request reference, and associations to ClaimEntity and PartyEntity.

FeePaymentRepository — Spring Data JPA repository with a findByRequestReference lookup method.

PaymentCallBackController — POST /service-request-update endpoint that receives service request update callbacks from the payments service and delegates to PaymentService.

ServiceRequestUpdate — Deserialisation model for the incoming payment callback payload.

Payment — Nested payment details model within ServiceRequestUpdate.

PaymentStatus enum — PAID, NOT_PAID, PARTIALLY_PAID with case-insensitive string lookup.  (NOTE THIS IS NOT DIRECTLY USED YET)

createServiceRequest now saves a new FeePaymentEntity after a successful service request is created.

New processPaymentResponse method looks up an existing FeePaymentEntity by request reference and updates its payment status.

New WireMock stub (payment-success.json) simulates a successful payment service request and fires a webhook callback back to the application.

### Testing done

Local round trip testing using a Docker based Wiremock implementation of the payment callback

image


src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
```diff
@@ -45,7 +46,7 @@ public PcsCaseEntity createCase(long caseReference,
         return pcsCaseRepository.save(pcsCaseEntity);
     }
 
-    public void createMainClaimOnCase(long caseReference, PCSCase pcsCase) {
+    public PartyEntity createMainClaimOnCase(long caseReference, PCSCase pcsCase) {
```
scottstewart-sl: This doesn't feel right for a createClaim method to return a PartyEntity

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/model/PaymentStatus.java
```diff
@@ -0,0 +1,26 @@
...
+    public static PaymentStatus fromValue(String value) {
+        for (PaymentStatus status : values()) {
+            if (status.value.equalsIgnoreCase(value)) {
+                return status;
+            }
+        }
+        throw new IllegalArgumentException("Unknown PaymentStatus value: " + value);
+    }
+
+    PaymentStatus(String s) {
```
scottstewart-sl: Minor - this `s` could have a better name. Is this, (and the field), the `label`? If you used that then that name would be more consistent with a lot of the other enums used in CCD

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentService.java
```diff
@@ -65,9 +78,59 @@ public PaymentServiceResponse createServiceRequest(
...
+            feePaymentRepository.save(feePaymentEntity);
+        }
+    }
+
+    private ClaimPartyEntity retrieveClaimPartyEntity(ClaimEntity claimEntity, String responsibleParty) {
+        return claimEntity.getClaimParties()
+            .stream()
+            .filter(party -> responsibleParty.equals(party.getParty().getOrgName()))
+            .findFirst()
+            .orElseThrow(() -> new IllegalStateException(MATCHING_PARTY_ENTITY_NOT_FOUND));
```
scottstewart-sl: Maybe use `PartyNotFoundException` for this? Or do you want to highlight the fact that the state is wrong if the party isn't found?

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentService.java
```diff
@@ -65,9 +78,59 @@ public PaymentServiceResponse createServiceRequest(
...
+    @Transactional
+    public void saveNewFeePayment(String caseReference, ClaimEntity claimEntity, ClaimPartyEntity claimParty,
+                                   FeeDto feeDto, String serviceRequestReference) {
+        log.info("Saving New Fee Payment for the case: {} with serviceRequestReference: {}", caseReference,
+                 serviceRequestReference);
+        FeePaymentEntity feePaymentEntity = FeePaymentEntity.builder()
+            .claim(claimEntity)
+            .requestReference(serviceRequestReference)
+            .amount(feeDto.getCalculatedAmount())
+            //.party(claimParty.getParty())
```
scottstewart-sl: Minor - commented line

src/main/resources/db/migration/V079__fee_payment.sql
```diff
@@ -0,0 +1,16 @@
...
+  party_id                      UUID,
+  request_date                  TIMESTAMP NOT NULL,
+  request_reference             VARCHAR(255),
+  external_reference            VARCHAR(255),
+  amount                        NUMERIC(19, 2),
+  payment_status                VARCHAR(50),
+
+  CONSTRAINT pk_fee_payment PRIMARY KEY (id),
+  CONSTRAINT fk_fee_payment_claim  FOREIGN KEY (claim_id)  REFERENCES claim (id)
+  --CONSTRAINT fk_fee_payment_party  FOREIGN KEY (party_id)  REFERENCES party (id)
```
scottstewart-sl: Minor - commented line

src/test/resources/wiremock/payment-success.json
```diff
@@ -0,0 +1,34 @@
+{
```
scottstewart-sl, line 1: Is this used in any tests? If it's just for local dev, what do you think about just putting it in a top level `wiremock` folder, like we have in pcs-frontend?

https://github.com/hmcts/pcs-frontend/tree/HDPI-4267_apply_something_else/wiremock

src/main/resources/db/migration/V079__fee_payment.sql
tvr-hmcts: Yep - deliberate at the moment.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
tvr-hmcts: There is a bug in this space.   Yes I thought the same to be honest but the refactor was much larger I think when I originally played it through.   The responsibleParty is the issue.

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentService.java
tvr-hmcts: Was a reminder for myself but I will need to apply it later.

src/test/resources/wiremock/payment-success.json
tvr-hmcts: It is just for local development though.   Placing it there would enable it whenever that label is applied within the pipeline no?   That maybe beyond the need and mask other behaviours if it is there.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
scottstewart-sl: It looks like you are only using the PartyEntity to get the `orgName` so you can put it in the payment request. If that is the case, then couldn't you just get that directly from the case data, even if you have to apply the fallback name logic?

src/test/resources/wiremock/payment-success.json
scottstewart-sl: No, I don't think so. The default mappings applied in the preview environment come from the [values.wiremock.preview.template.yaml](https://github.com/hmcts/pcs-api/blob/master/charts/pcs-api/values.wiremock.preview.template.yaml#L16) file, with any extra mappings POSTed on the fly depending on developer/tester need.

Then we can, (in a later PR), add a docker compose file for the local dependencies that includes spinning up a wiremock server with those mappings, like in the pcs-frontend: https://github.com/hmcts/pcs-frontend/blob/HDPI-4266_help_with_fees/docker-compose-dependencies-only.yml#L27-L32. We could also, (in a later PR), put the FeeRegister mapping in there and then remove the FakeFeeService code as well.

tvr-hmcts: Interesting.  Yes that could possible then.  Due to the nature of this ticket though I think the movement will need to be done in a later ticket also.   There is a bit of a time lag in this ticket.   It is associated with a spike and integration issues.

charts/pcs-api/values.preview.template.yaml
```diff
@@ -25,6 +25,7 @@ java:
         - name: pcs-prd-admin-password
           alias: PCS_PRD_ADMIN_PASSWORD
   environment:
+    CHANGE_ID: ${CHANGE_ID}
```
linusnorton: whats this for?

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentService.java
```diff
@@ -65,9 +79,58 @@ public PaymentServiceResponse createServiceRequest(
...
+    private ClaimPartyEntity retrieveClaimPartyEntity(ClaimEntity claimEntity, String responsibleParty) {
+        return claimEntity.getClaimParties()
+            .stream()
+            .filter(party -> responsibleParty.equals(party.getParty().getOrgName()))
+            .findFirst()
+            .orElseThrow(() -> new PartyNotFoundException(MATCHING_PARTY_ENTITY_NOT_FOUND));
+    }
+
+    @Transactional
+    public void saveNewFeePayment(String caseReference, ClaimEntity claimEntity, ClaimPartyEntity claimParty,
```
scottstewart-sl: Is `claimParty` used anywhere?

src/integrationTest/java/uk/gov/hmcts/reform/pcs/controllers/PaymentCallBackControllerIT.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    PcsCaseEntity establishTestCase(long caseReference) {
+        return caseCreationHelper
+            .createTestCaseWithParty(caseReference, null, PartyRole.DEFENDANT);
+    }
+
+    void establishFeePayment(PcsCaseEntity pcsCaseEntity, String serviceCaseReference, FeeDto feeDto) {
+        ClaimEntity claimEntity = pcsCaseEntity.getClaims().getFirst();
+        // ClaimPartyEntity claimPartyEntity = claimEntity.getClaimParties().getFirst();
```
linusnorton: clean up?

docker-compose-wiremock.yml
```diff
@@ -0,0 +1,9 @@
+services:
```
scottstewart-sl, line 1: Minor - I think it would be a bit more consistent to call this `docker-compose-wirewmock.yml`. I don't think it needs the "local" in the name as that's implied by the fact that it is being run in docker.

That naming convention is more aligned with pcs-frontend which already has docker compose files in it.

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentService.java
tvr-hmcts: It will be.   I removed the linking to the Party in this iteration

charts/pcs-api/values.preview.template.yaml
tvr-hmcts: Legacy testing.  Removing.

src/main/resources/db/migration/V081__fee_payment.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE fee_payment (
+  id                            UUID NOT NULL,
+  claim_id                      UUID NOT NULL,
+  party_id                      UUID,
```
saurabhvalsangkar-hmcts: This field is not mapped on fee_payment entity

charts/pcs-api/values.yaml
```diff
@@ -42,6 +42,7 @@ java:
     LOCATION_REF_URL: "http://rd-location-ref-api-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal"
     FEES_REGISTER_API_URL: "http://fees-register-api-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal"
     PAYMENT_API_URL: "http://payment-api-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal"
+    PAY_CALLBACK_URL: "http://pcs-api-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal/service-request-update"
```
scottstewart-sl: Minor - not a biggie but all the other url envs here are just to the host, and then the path is presumably appended in code. That means that if the request path changes, it wouldn't need to be updated everwhere  this has been overridden.

Also it could be called PCS_API_URL to be more generic and re-usable by anything else that wants a callback.

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/endpoint/PaymentCallBackController.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public static final String SERVICE_AUTHORIZATION = "ServiceAuthorization";
+    public static final String PAYMENT_UPDATE_PATH = "/payment-update";
+
+    private final PaymentService paymentService;
+
+    @PutMapping(path = PAYMENT_UPDATE_PATH, consumes = APPLICATION_JSON_VALUE)
+    public void processPaymentCallback(
+        @RequestHeader(value = HttpHeaders.AUTHORIZATION, required = false) String authorisation,
+        @RequestHeader(value = SERVICE_AUTHORIZATION, required = false) String s2sToken,
```
scottstewart-sl: Why are these required = false? I think everywhere else in the code they are left to the default which is true

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentService.java
```diff
@@ -9,18 +10,32 @@
...
 import uk.gov.hmcts.reform.pcs.idam.IdamService;
 
+import java.util.Optional;
+
 @Slf4j
 @Service
 @RequiredArgsConstructor
 public class PaymentService {
 
+    public static final String MATCHING_PARTY_ENTITY_NOT_FOUND = "Matching PartyEntity not found";
```
scottstewart-sl: Minor - can this be private?

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentService.java
scottstewart-sl: It would be cleaner to just add it back in when you need it. Otherwise there's a chance that if the planned changes don't happen, we're left with an unused parameter.

charts/pcs-api/values.yaml
scottstewart-sl: In fact this has already happened - it looks like `/service-request-update` has now become `/payment-update` but this one hasn't been updated.

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/endpoint/PaymentCallBackController.java
tvr-hmcts: Yep.   Applied initially will change back.

charts/pcs-api/values.yaml
tvr-hmcts: Good spot thanks.   I missed on the update.

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentService.java
tvr-hmcts: Was being used within a test.  Reverting.

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentService.java
tvr-hmcts: Ok fair.

scottstewart-sl (approved):

src/integrationTest/java/uk/gov/hmcts/reform/pcs/controllers/PaymentCallBackControllerIT.java
tvr-hmcts: This is gone now.

docker-compose-wiremock.yml
andrewdami11, line 1: Did we mean `wiremock`, instead of wirewmock, for this file's name?

src/main/resources/db/migration/V082__fee_payment.sql
```diff
@@ -0,0 +1,15 @@
...
+  request_reference             VARCHAR(255),
+  external_reference            VARCHAR(255),
+  amount                        NUMERIC(19, 2),
+  payment_status                VARCHAR(50),
+
+  CONSTRAINT pk_fee_payment PRIMARY KEY (id),
+  CONSTRAINT fk_fee_payment_claim  FOREIGN KEY (claim_id)  REFERENCES claim (id)
+);
+
+CREATE INDEX idx_fee_payment_request_reference ON fee_payment (request_reference);
```
andrewdami11, line 15: If `request_reference` is a business key (as `findByRequestReference` suggests), would a `UNIQUE` index be safer than a non-unique one, so duplicate service-request references cannot be inserted?

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/endpoint/PaymentCallBackController.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public static final String SERVICE_AUTHORIZATION = "ServiceAuthorization";
+    public static final String PAYMENT_UPDATE_PATH = "/payment-update";
+
+    private final PaymentService paymentService;
+
+    @PutMapping(path = PAYMENT_UPDATE_PATH, consumes = APPLICATION_JSON_VALUE)
+    public void processPaymentCallback(
+        @RequestHeader(value = AUTHORIZATION) String authorisation,
+        @RequestHeader(value = SERVICE_AUTHORIZATION) String s2sToken,
```
benouaer, line 28: we should probably validate the s2sToken

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentService.java
```diff
@@ -65,9 +76,49 @@ public PaymentServiceResponse createServiceRequest(
...
+
+        return paymentServiceResponse;
+    }
+
+    @Transactional
+    public void processPaymentResponse(ServiceRequestUpdate serviceRequestUpdate) {
+        log.info("ServiceRequestUpdate status: {}", serviceRequestUpdate.getServiceRequestStatus());
+        Optional<FeePaymentEntity> byCaseReference = feePaymentRepository
+            .findByRequestReference(serviceRequestUpdate.getServiceRequestReference());
+        if (byCaseReference.isPresent()) {
```
adusumillipraveen, line 97: I think we should error / log it if we get to such scneario.

src/main/resources/db/migration/V082__fee_payment.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE fee_payment (
+  id                            UUID NOT NULL,
+  claim_id                      UUID NOT NULL,
+  party_id                      UUID,
```
andrewdami11, line 4: The table includes `party_id` but the JPA entity does not map it yet, and there is no FK to party. Should this be handled?

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentService.java
```diff
@@ -65,9 +76,49 @@ public PaymentServiceResponse createServiceRequest(
...
+                          paymentServiceResponse.getServiceRequestReference());
+
+        return paymentServiceResponse;
+    }
+
+    @Transactional
+    public void processPaymentResponse(ServiceRequestUpdate serviceRequestUpdate) {
+        log.info("ServiceRequestUpdate status: {}", serviceRequestUpdate.getServiceRequestStatus());
+        Optional<FeePaymentEntity> byCaseReference = feePaymentRepository
+            .findByRequestReference(serviceRequestUpdate.getServiceRequestReference());
```
benouaer, line 96: is the serviceRequestReference the same as the caseReference?

src/main/resources/db/migration/V082__fee_payment.sql
tvr-hmcts: It will be.  There was a seperate bug in this space which I hope has beed addressed in this PR but the linking through to the Party has been left out for now.   Ive kept it in here though to save on another flyway as this is the table structure from the DM

src/test/java/uk/gov/hmcts/reform/pcs/ccd/event/ResumePossessionClaimTest.java
```diff
@@ -205,6 +208,9 @@ void setUp() {
         when(securityContextService.getCurrentUserDetails()).thenReturn(userDetails);
         when(userDetails.getUid()).thenReturn(USER_ID.toString());
 
+        PartyEntity claimant = mock(PartyEntity.class);
```
benouaer: should we building the claimant object rather than mocking it, given that it's not a service?

src/test/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentServiceTest.java
```diff
@@ -13,100 +13,265 @@
...
-        String systemToken = "Bearer sys-token";
-
-        FeeDto mappedFee = FeeDto.builder()
-            .calculatedAmount(new BigDecimal("808.00"))
-            .code("FEE0412")
-            .version("4")
-            .volume(volume)
-            .build();
+        // Given
+        FeeDetails feeDetails = mock(FeeDetails.class);
```
benouaer: can this object be populated rather than mocked?

src/test/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentServiceTest.java
```diff
@@ -13,100 +13,265 @@
...
+
+        // Then
+        verify(feePaymentRepository).findByRequestReference(requestReference);
+        verify(feePaymentRepository, org.mockito.Mockito.never()).save(any(FeePaymentEntity.class));
+    }
+
+    @Test
+    void shouldSaveNewFeePaymentWithExpectedFields() {
+        // Given
+        ClaimEntity claimEntity = mock(ClaimEntity.class);
```
benouaer: can this be constructed rather than mocked?

src/test/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentServiceTest.java
```diff
@@ -13,100 +13,265 @@
...
-        );
+        when(idamService.getSystemUserAuthorisation()).thenReturn(SYSTEM_TOKEN);
+    }
 
+    private PcsCaseEntity setupPcsCase(ClaimPartyEntity claimPartyEntity) {
+        ClaimEntity claimEntity = mock(ClaimEntity.class);
+        lenient().when(claimEntity.getClaimParties()).thenReturn(List.of(claimPartyEntity));
+        PcsCaseEntity pcsCaseEntity = mock(PcsCaseEntity.class);
+        when(pcsCaseEntity.getClaims()).thenReturn(List.of(claimEntity));
+        return pcsCaseEntity;
```
benouaer: can we construct these entities rather than mock them

benouaer (approved):

benouaer (approved):

docker-compose-wiremock.yml
tvr-hmcts: Kenesis keyboard fail.   (I'm blaming the tool)

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/endpoint/PaymentCallBackController.java
tvr-hmcts: Mentioned similar in another PR review.   Wondering if there should be a filter actually for this kind of thing.

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentService.java
tvr-hmcts: It can be thought of as a reference to the actual payment 'session' and it is returned from CCPay on the Service Request.   We call it 'request reference' with the DM.   It is the 'key' for if we want to look up the status of the payment.

src/main/resources/db/migration/V082__fee_payment.sql
tvr-hmcts: Possibly.  Good thinking.  However, based on what I've read and the use cases I'm not completely sure at this time.

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentService.java
tvr-hmcts: Ill add an error log with the else for now.

There is a seperate ticket for the error scenarios from Payment -> https://tools.hmcts.net/jira/browse/HDPI-5501

adusumillipraveen (approved):

toby-plunkett (approved):

PR #1624, opened 2026-04-07
Enable Case Flags Functionality

aschalew-at:
### Jira link

See [HDPI-3503](https://tools.hmcts.net/jira/browse/HDPI-3503)

  * [ ] Yes
  * [ ] No

### Checklist


src/main/resources/db/migration/V085__create_case_flags_related_tables.sql
```diff
@@ -0,0 +1,33 @@
+CREATE TABLE flag_details (
```
gmmagruder: I don't see these changes in the data model. Is Saurabh aware of these changes?

src/main/resources/application.yaml
```diff
@@ -24,6 +24,7 @@ globalsearch:
 
 hmcts:
   hmctsOrgId: AAA3
+  alternativeHmctsOrgId: AAA7
```
tvr-hmcts: Not sure on the name - should it be related to where it is used?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -0,0 +1,190 @@
...
+        return mergedPartyEntities;
+    }
+
+    private void mergePartyFlagGroup(Flags incomingPartyFlags, List<FlagDetailsEntity> existingFlags,
+                                     PartyEntity partyEntity, PcsCaseEntity pcsCaseEntity) {
+        Map<UUID, FlagDetailsEntity> existingFlagDetailsMap =
+            existingFlags.stream()
+                .collect(Collectors.toMap(FlagDetailsEntity::getId, Function.identity()));
+
+        List<String> existingFlagPathIds = getExistingPathIds(existingFlagDetailsMap);
```
tvr-hmcts: There are more possibilities with this getExistingPathIds(...) invocation but here at least it can go inside the if statement below and also the existingFlagDetailsMap in this case as I do not see why they need to be processed if the incomingPartyFlags.getDetails() were null etc

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -22,28 +23,87 @@
...
-    public void mergeCaseFlags(Flags incomingCaseFlags, PcsCaseEntity pcsCaseEntity) {
 
-        Map<UUID, FlagDetailsEntity> existingFlagDetails =
+    public void mergeCaseFlags(Flags incomingCaseFlags, PcsCaseEntity pcsCaseEntity, List<ListValue<Party>> parties) {
+        Map<UUID, FlagDetailsEntity> existingFlagDetailsMap =
             pcsCaseEntity.getCaseFlags().stream()
                 .collect(Collectors.toMap(FlagDetailsEntity::getId, Function.identity()));
 
-        List<FlagDetailsEntity> mergedFlagDetails = new ArrayList<>();
+        List<String> existingFlagPathIds = getExistingPathIds(existingFlagDetailsMap);
```
tvr-hmcts: Could the existingFlagPathIds be returned in a HashSet rather than creating a HashSet from it in line 164?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/RefDataFlagsRepository.java
```diff
@@ -1,12 +1,13 @@
...
 import uk.gov.hmcts.reform.pcs.ccd.entity.RefDataFlagsEntity;
 
-import java.util.Optional;
 import java.util.UUID;
 
+@Repository
 public interface RefDataFlagsRepository extends JpaRepository<RefDataFlagsEntity, UUID> {
 
-    Optional<RefDataFlagsEntity> findByFlagCode(String flagCode);
+    RefDataFlagsEntity findByFlagCode(String flagCode);
```
tvr-hmcts: Is this guaranteed?  I see the change from the Optional return

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -22,28 +23,87 @@
...
+                                                     List<String> existingFlagPathIds, PartyEntity partyEntity) {
+
+        List<FlagDetailsEntity> mergedFlagDetails = new ArrayList<>();
 
         for (ListValue<FlagDetail> incomingFlagDetailListValue : incomingCaseFlags.getDetails()) {
             FlagDetail incomingFlagDetail = incomingFlagDetailListValue.getValue();
 
-            RefDataFlagsEntity refDataFlagsEntity = refDataFlagsRepository.findByFlagCode((
-                incomingFlagDetail.getFlagCode())).orElse(null);
+            RefDataFlagsEntity refDataFlagsEntity = refDataFlagsRepository.findByFlagCode(
```
tvr-hmcts: The RefDataFlagsEntity is different to what else happens in this loop.   Can it be extracted to another method ... return the refDataFlagsEntity ... hold that within a collection outside of this loop and then at the end of the loop call the refDataFlagsRepository.saveAll(...)

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -22,28 +23,87 @@
...
+                .collect(Collectors.toMap(PartyEntity::getId,
+                                          Function.identity()
+                ));
+        List<PartyEntity> mergedPartyEntities = new ArrayList<>();
+
+        for (ListValue<Party> incomingPartyValue : incomingParties) {
+            Party incomingParty = incomingPartyValue.getValue();
 
-        List<String> existingFlagPathIds = getExistingPathIds(existingFlagDetails);
+            PartyEntity partyEntity = existingPartiesMap.get(UUID.fromString(incomingPartyValue.getId()));
```
tvr-hmcts, line 53: Is the partyEntity guaranteed here?

tvr-hmcts (commented):
Some thoughts.

src/main/resources/db/migration/V085__create_case_flags_related_tables.sql
aschalew-at: Table name is changed to case_flag and the table details are passed to the data team to be added in data model.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
aschalew-at: I included both in the if block

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
aschalew-at: getExistingPathIds() returns now Set<String>

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/RefDataFlagsRepository.java
aschalew-at: Reverted back to Optional<....

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
aschalew-at: Refactored code.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
aschalew-at: Added null check for partyEntity

src/main/resources/application.yaml
toby-plunkett: +1

toby-plunkett (approved):
Just that one name change Trevor mentioned

toby-plunkett:
Could you add a description to both this PR and your other one?

 I'm also seeing a lot of similar methods across both PRs, e.g mergeCaseFlags, is this PR intending to override those differences? If so, add a comment on the old PR so we know not to review the code which will be replaced.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/BaseCaseFlag.java
```diff
@@ -44,7 +44,11 @@ public abstract class BaseCaseFlag {
 
     @OneToMany(mappedBy = "caseFlagEntity", cascade = ALL, orphanRemoval = true)
     @Builder.Default
-    private List<FlagPathEntity> paths = new ArrayList<>();
+    private List<FlagPathEntity> caseFlagPaths = new ArrayList<>();
+
+    @OneToMany(mappedBy = "casePartyFlagEntity", cascade = ALL, orphanRemoval = true)
+    @Builder.Default
+    private List<FlagPathEntity> casePartyFlagPaths = new ArrayList<>();
```
benouaer: why are there two separate fields for paths?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -5,32 +5,130 @@
...
+            mergedPartyEntities.add(partyEntity);
+        }
+
+        return mergedPartyEntities;
+    }
+
+    private void mergePartyFlagGroup(Flags incomingPartyFlags, List<CasePartyFlagEntity> existingFlags,
+                                     PartyEntity partyEntity, PcsCaseEntity pcsCaseEntity, String flow) {
+
+        if (incomingPartyFlags.getDetails() != null && !incomingPartyFlags.getDetails().isEmpty()) {
```
benouaer: this check is already being done on lines 82 and 83

src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/CasePartyFlagEntity.java
```diff
@@ -0,0 +1,28 @@
...
+
+
+@Entity
+@Table(name = "case_party_flag")
+@Setter
+@Getter
+@AllArgsConstructor
+@NoArgsConstructor
+@Builder
+public class CasePartyFlagEntity extends BaseCaseFlag {
```
benouaer, line 23: can we call it PartyFlagEntity? it's a bit confusing when you add case to it

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -5,32 +5,130 @@
...
+
+            pcsCaseEntity.getCaseFlags().clear();
+            pcsCaseEntity.getCaseFlags().addAll(mergedCaseFlagDetails);
+        }
+
+        if (parties != null) {
+            List<PartyEntity> mergedPartyEntities = mergePartyFlags(parties, pcsCaseEntity, flow);
+
+            pcsCaseEntity.getParties().clear();
+            pcsCaseEntity.getParties().addAll(mergedPartyEntities);
```
benouaer: this doesn't look right, we're not changing the parties, we're just changing party flags
so it doesn't look to me like we should be updating the list of parties
we should be looping through the parties, and for each party updating the flags

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -5,32 +5,130 @@
...
+
+        return mergedPartyEntities;
+    }
+
+    private void mergePartyFlagGroup(Flags incomingPartyFlags, List<CasePartyFlagEntity> existingFlags,
+                                     PartyEntity partyEntity, PcsCaseEntity pcsCaseEntity, String flow) {
+
+        if (incomingPartyFlags.getDetails() != null && !incomingPartyFlags.getDetails().isEmpty()) {
+
+            Map<UUID, BaseCaseFlag> existingCaseFlagEntityMap =
```
benouaer: these are not case flags right, they're party flags no?

tvr-hmcts:
Does this need to be updated / aligned with the target if the target has moved.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/Party.java
```diff
@@ -36,4 +39,11 @@ public class Party {
 
     private LocalDate dateOfBirth;
 
+    @CCD(
+        access = {InternalCaseFlagAccess.class},
+        label = "Party Flags",
+        retainHiddenValue = true
+    )
+    private Flags respondentFlags;
```
gmmagruder: Wouldn't it make more sense to call them defendantFlags or just flags if this could also be used by claimants in future?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/party/PartyEntity.java
```diff
@@ -111,4 +112,10 @@ public class PartyEntity {
     @JsonManagedReference
     private List<ClaimPartyLegalRepresentativeEntity> claimPartyLegalRepresentativeList = new ArrayList<>();
 
+    @OneToMany(mappedBy = "party",
+        cascade = ALL,
+        orphanRemoval = true)
+    @Builder.Default
+    private List<CasePartyFlagEntity> respondentFlags = new ArrayList<>();
```
gmmagruder: Same as above

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/ManageFlags.java
```diff
@@ -34,6 +35,11 @@ public void configureDecentralised(DecentralisedConfigBuilder<PCSCase, State, Us
             .page("caseworkerCaseFlag")
             .pageLabel("Case Flags")
             .optional(PCSCase::getCaseFlags, ShowConditions.NEVER_SHOW, true, true)
+            .optional(PCSCase::getParties, ShowConditions.NEVER_SHOW, true, true)
+            .page("respondentFlags")
+            .list(PCSCase::getParties, ShowConditions.NEVER_SHOW)
```
gmmagruder: You can use getAllDefendants so you don't have to fetch the claimant as well

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/Party.java
aschalew-at: The how-to-guide (https://tools.hmcts.net/confluence/pages/viewpage.action?pageId=1572570790&spaceKey=RCCD&title=How%2BTo%2BGuide%2B-%2BCase%2BFlags ) suggests respondentFlags

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
aschalew-at: Renamed to existingPartyFlagEntityMap

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/ManageFlags.java
aschalew-at: Called getAllDefendants() instead of getParties()

src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/party/PartyEntity.java
aschalew-at: Refactored to defendantFlags.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/Party.java
aschalew-at: Refactored to defendantFlags.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -30,12 +34,66 @@ public class CaseFlagService {
 
     private FlagRefDataRepository flagRefDataRepository;
 
-    public List<BaseCaseFlag> mergeCaseFlags(Flags incomingCaseFlags, PcsCaseEntity pcsCaseEntity, String flow) {
-        Map<UUID, BaseCaseFlag> existingFlagEntities =
-            pcsCaseEntity.getCaseFlags().stream()
+    public List<CaseFlagEntity> mergeCaseFlags(Flags incomingCaseFlags, PcsCaseEntity pcsCaseEntity, String flow) {
+
+        if (incomingCaseFlags != null && incomingCaseFlags.getDetails() != null) {
```
benouaer: this check is already being done in the pcsCaseService

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -30,12 +34,66 @@ public class CaseFlagService {
...
+                    PartyEntity::getId,
+                    Function.identity()
+                ));
+
+            for (ListValue<Party> incomingPartyValue : incomingParties) {
+                Party incomingParty = incomingPartyValue.getValue();
+
+                PartyEntity partyEntity = existingPartiesMap.get(UUID.fromString(incomingPartyValue.getId()));
+
+                if (partyEntity != null && incomingParty.getDefendantFlags() != null
```
benouaer: partyEntity cannot be null because we just populated the map by going through the list of existing parties on the case

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -30,12 +34,66 @@ public class CaseFlagService {
...
+        List<CasePartyFlagEntity> mergedCasePartyFlags = mergeFlagDetails(
+            existingPartyFlagEntityMap, incomingPartyFlags, pcsCaseEntity,
+            partyEntity, flow, CasePartyFlagEntity::new);
+
+        partyEntity.getDefendantFlags().clear();
+        partyEntity.getDefendantFlags().addAll(mergedCasePartyFlags);
+    }
+
+    private <T extends BaseCaseFlag> List<T>  mergeFlagDetails(
+                        Map<UUID, T> existingCaseFlagEntitiesMap, Flags incomingCaseFlags, PcsCaseEntity pcsCaseEntity,
```
benouaer: can we rename existingCaseFlagEntitiesMap and incomingCaseFlags, because they can refer to both case and party flags

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
```diff
@@ -69,11 +68,14 @@ public void patchCaseFlags(long caseReference, PCSCase pcsCase, String flow) {
         }
         PcsCaseEntity pcsCaseEntity = loadCase(caseReference);
 
-        if (pcsCase.getCaseFlags() != null && pcsCase.getCaseFlags().getDetails() != null) {
-            List<BaseCaseFlag> mergedFlagDetails = caseFlagService.mergeCaseFlags(pcsCase.getCaseFlags(),
-                                                                                  pcsCaseEntity, flow);
+        if (pcsCase.getCaseFlags() != null && pcsCase.getCaseFlags().getDetails() != null
```
benouaer: I think we should separate this if statement into two, one for caseFlags and one for partyFlags, because the merge operations are independent

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseFlagsView.java
```diff
@@ -0,0 +1,148 @@
...
+import uk.gov.hmcts.reform.pcs.ccd.domain.PCSCase;
+import uk.gov.hmcts.reform.pcs.ccd.domain.Party;
+import uk.gov.hmcts.reform.pcs.ccd.entity.PcsCaseEntity;
+import uk.gov.hmcts.reform.pcs.ccd.entity.BaseCaseFlag;
+import uk.gov.hmcts.reform.pcs.ccd.util.YesOrNoConverter;
+import uk.gov.hmcts.reform.pcs.ccd.entity.party.PartyEntity;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.ArrayList;
```
tvr-hmcts: Not sure how that snuck through as it is not needed - thought checkstyle would have caught it.

src/test/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseFlagsViewTest.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.List;
+import java.util.Set;
+import java.util.UUID;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+@ExtendWith(MockitoExtension.class)
+class CaseFlagsViewTest {
```
tvr-hmcts, line 24: Could there be some additional assertions to ensure that the logic within mapDefendantFlags(...) is covered.   IE: that the Flags have the correct party names etc within them

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagServiceTest.java
```diff
@@ -0,0 +1,315 @@
...
+
+        underTest.mergePartyFlags(parties, pcsCaseEntity, EventFlow.CREATE.name());
+
+        assertEquals(1, pcsCaseEntity.getParties().size());
+        PartyEntity savedParty = pcsCaseEntity.getParties().iterator().next();
+
+        assertNotNull(savedParty.getDefendantFlags());
+        assertEquals(1, savedParty.getDefendantFlags().size());
+
+        BaseCaseFlag savedFlags = savedParty.getDefendantFlags().getFirst();
```
tvr-hmcts, line 143: Does this need completing?

Assertions on FlagComment

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagServiceTest.java
```diff
@@ -0,0 +1,315 @@
...
+        // When
+        underTest.mergePartyFlags(incomingParties, pcsCaseEntity, EventFlow.UPDATE.name());
+
+        // Then
+        assertEquals(1, pcsCaseEntity.getParties().size());
+        PartyEntity updatedParty = pcsCaseEntity.getParties().iterator().next();
+
+        assertNotNull(updatedParty.getDefendantFlags());
+        assertEquals(1, updatedParty.getDefendantFlags().size());
+
```
tvr-hmcts, line 206: assertThat(...) for the FlagUpdateComment as this is an UPDATE

It maybe worth having a paramaterised test based on the EventFlow so to assert on the FlagUpdateComment when under "Inactive" also.

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagServiceTest.java
```diff
@@ -0,0 +1,315 @@
...
+        // When
+        List<CaseFlagEntity> savedFlags = underTest.mergeCaseFlags(incomingFlags,
+                                                                   pcsCaseEntity, EventFlow.CREATE.name());
+
+        // Then
+        String savedPaths = savedFlags.getFirst().getPaths();
+        assertEquals("Active", savedFlags.getFirst().getDefaultStatus());
+        assertEquals(1, savedFlags.size());
+        assertThat(savedPaths).isNull();
```
tvr-hmcts: Can there be an assertion that the PcsCase has been set when under this flow.

src/test/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/InternalCaseFlagAccessTest.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    @Test
+    void shouldGrantInternalCaseFlagAccess() {
+
+        // When
+        SetMultimap<HasRole, Permission> grants = underTest.getGrants();
+
+        // Then
+        assertThat(grants.asMap()).contains(entry(PCS_CASE_WORKER, Permission.CRU));
```
tvr-hmcts: Pedantic I know but this could assert on all of them and the size of the Set

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/InternalCaseFlagAccess.java
```diff
@@ -0,0 +1,29 @@
...
+import static uk.gov.hmcts.reform.pcs.ccd.accesscontrol.UserRole.WLU_ADMIN;
+import static uk.gov.hmcts.reform.pcs.ccd.accesscontrol.UserRole.BAILIFF_ADMIN;
+
+
+public class InternalCaseFlagAccess implements HasAccessControl {
+
+    @Override
+    public SetMultimap<HasRole, Permission> getGrants() {
+        SetMultimap<HasRole, Permission> grants = HashMultimap.create();
+        grants.putAll(PCS_CASE_WORKER, Permission.CRU);
```
benouaer: remove this role, it's not needed anymore

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/CreateFlags.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    public void configureDecentralised(DecentralisedConfigBuilder<PCSCase, State, UserRole> configBuilder) {
+        new PageBuilder(configBuilder
+                .decentralisedEvent(EventId.createFlags.name(), this::submit)
+                .forState(State.PENDING_CASE_ISSUED)
+                .name("Create case flags")
+                .description("To create flags")
+                .showSummary()
+                .grant(Permission.CRU,
+                       UserRole.PCS_CASE_WORKER,
```
benouaer: we don't need this role anymore

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/ManageFlags.java
```diff
@@ -0,0 +1,60 @@
...
+    @Override
+    public void configureDecentralised(DecentralisedConfigBuilder<PCSCase, State, UserRole> configBuilder) {
+        new PageBuilder(configBuilder
+                            .decentralisedEvent(EventId.amendFlags.name(), this::submit)
+                            .forState(State.PENDING_CASE_ISSUED)
+                            .name("Manage case flags")
+                            .description("To manage flags")
+                            .showSummary()
+                            .grant(Permission.CRU,
+                                   UserRole.PCS_CASE_WORKER,
```
benouaer: we don't need this role anymore

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/ManageFlags.java
```diff
@@ -0,0 +1,60 @@
...
+@Slf4j
+@AllArgsConstructor
+public class ManageFlags implements CCDConfig<PCSCase, State, UserRole> {
+    private  final PcsCaseService pcsCaseService;
+
+    @Override
+    public void configureDecentralised(DecentralisedConfigBuilder<PCSCase, State, UserRole> configBuilder) {
+        new PageBuilder(configBuilder
+                            .decentralisedEvent(EventId.amendFlags.name(), this::submit)
+                            .forState(State.PENDING_CASE_ISSUED)
```
benouaer, line 30: is the event only available when the state is State.PENDING_CASE_ISSUED, or is it also available for State.CASE_ISSUED?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -0,0 +1,176 @@
...
+                pcsCaseEntity.getCaseFlags().stream()
+                    .collect(Collectors.toMap(BaseCaseFlag::getId, Function.identity()));
+
+        return mergeFlagDetails(existingCaseFlagEntitiesMap, incomingCaseFlags, pcsCaseEntity,null, flow,
+                        CaseFlagEntity::new);
+    }
+
+    public void mergePartyFlags(List<ListValue<Party>> incomingParties, PcsCaseEntity pcsCaseEntity,
+                                             String flow) {
+        if (incomingParties != null) {
```
benouaer: incomingParties cannot be null, because you're already checking for that in PcsCaseService

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -0,0 +1,176 @@
...
+                flagEntity = flagEntitySupplier.get();
+            }
+
+            if (flagEntity instanceof CasePartyFlagEntity casePartyFlagEntity) {
+                casePartyFlagEntity.setParty(partyEntity);
+            } else if (flagEntity instanceof CaseFlagEntity caseFlagEntity) {
+                caseFlagEntity.setPcsCase(pcsCaseEntity);
+            }
+
+            if ("CREATE".equals(flow)) {
```
benouaer: not sure about this logic, wouldn't updateComment always go on updateComment?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -0,0 +1,176 @@
...
+            mergedFlagDetails.add(flagEntity);
+        }
+        flagRefDataRepository.saveAll(flagRefDataEntities);
+
+        return mergedFlagDetails;
+    }
+
+    private FlagRefDataEntity getRefDataEntity(Flags incomingCaseFlags, FlagDetail incomingFlagDetail) {
+
+        FlagRefDataEntity refDataFlagsEntity = flagRefDataRepository.findByFlagCode(
```
benouaer: I think this should be renamed flagRefdataEntity

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -0,0 +1,176 @@
...
+                        Map<UUID, T> existingFlagEntitiesMap, Flags incomingCaseFlags, PcsCaseEntity pcsCaseEntity,
+                        PartyEntity partyEntity, String flow, Supplier<T> flagEntitySupplier) {
+
+        List<T> mergedFlagDetails = new ArrayList<>();
+        Set<FlagRefDataEntity> flagRefDataEntities = new HashSet<>();
+
+        for (ListValue<FlagDetail> incomingFlagDetailListValue : incomingCaseFlags.getDetails()) {
+            FlagDetail incomingFlagDetail = incomingFlagDetailListValue.getValue();
+
+            FlagRefDataEntity flagRefDataEntity = getRefDataEntity(incomingCaseFlags, incomingFlagDetail);
```
benouaer: I think visibility should be calculated outside the loop and passed to this method if it's the same value for all incoming flags

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseFlagsView.java
```diff
@@ -0,0 +1,148 @@
...
+                       caseFlagEntity.getFlagRefData().getAvailableExternally()))
+                   .path(getPaths(caseFlagEntity.getPaths()))
+                   .build())
+                .build())
+            .toList();
+    }
+
+    private List<ListValue<String>> getPaths(String entityPaths) {
+
+        return entityPaths == null
```
benouaer: can this actually ever be null? is there a flag without paths? is the field not set to notNullable?

src/main/resources/db/migration/V100__create_case_flags_related_tables.sql
```diff
@@ -0,0 +1,48 @@
...
+                            other_description VARCHAR(50),
+                            other_description_cy VARCHAR(50),
+                            flag_comment VARCHAR(255),
+                            flag_comment_cy VARCHAR(255),
+                            flag_update_comment VARCHAR(255),
+                            flag_update_comment_cy VARCHAR(255),
+                            date_time_created TIMESTAMP,
+                            date_time_modified TIMESTAMP,
+                            status VARCHAR(50) NOT NULL,
+                            paths VARCHAR(255)
```
benouaer: can paths be null?

src/main/resources/application.yaml
```diff
@@ -208,3 +208,4 @@ payments:
 core_case_data:
   api:
     url: ${CCD_DATA_STORE_URL:localhost:4452}
+  defaultCaseApiUrl: http://localhost:3206
```
benouaer: this should not be needed, why can't the default be set in the same way as line 210 above

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -0,0 +1,176 @@
...
+
+    private void setFlagPath(FlagDetail incomingFlagDetail, BaseCaseFlag flagEntity) {
+
+        if (incomingFlagDetail.getPath() != null) {
+            StringBuilder paths = new StringBuilder();
+            incomingFlagDetail.getPath().stream()
+                .map(pathListValue -> pathListValue.getId() + CaseFlagsView.PATH_DELIMITER + pathListValue.getValue())
+                .forEach(path -> paths.append(path).append(CaseFlagsView.PATHS_DELIMITER));
+
+            paths.deleteCharAt(paths.lastIndexOf(CaseFlagsView.PATHS_DELIMITER));
```
benouaer: you can use a joining collector to avoid having to delete the last delimiter
`
String paths = incomingFlagDetail.getPath().stream()
                .map(pathListValue -> pathListValue.getId() + CaseFlagsView.PATH_DELIMITER + pathListValue.getValue())
                .collect(Collectors.joining(CaseFlagsView.PATHS_DELIMITER));
`

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -0,0 +1,176 @@
...
+            flagRefDataEntities.add(flagRefDataEntity);
+
+            T flagEntity = existingFlagEntitiesMap.remove(UUID.fromString(incomingFlagDetailListValue.getId()));
+
+
+            if (flagEntity == null) {
+                flagEntity = flagEntitySupplier.get();
+            }
+
+            if (flagEntity instanceof CasePartyFlagEntity casePartyFlagEntity) {
```
benouaer: you can avoid the instance check here by declaring an abstract setParentEntity method in BaseCaseFlag and then each child class implements it accordingly
```
================
BaseCaseFlag
abstract public void setParentEntity(PcsCaseEntity caseEntity, PartyEntity partyEntity)
================
CaseFlagEntity
    @Override
    public void setParentEntity(PcsCaseEntity caseEntity, PartyEntity parentEntity) {
        this.pcsCase = caseEntity;
    }
===================
CasePartyFlagEntity
@Override
    public void setParentEntity(PcsCaseEntity caseEntity, PartyEntity partyEntity) {
        this.party = parentEntity;
    }
```

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -0,0 +1,176 @@
...
+
+    private void mergePartyFlagGroup(Flags incomingPartyFlags, List<CasePartyFlagEntity> existingFlags,
+                                     PartyEntity partyEntity, PcsCaseEntity pcsCaseEntity, String flow) {
+
+        Map<UUID, CasePartyFlagEntity> existingPartyFlagEntityMap =
+            existingFlags.stream()
+                .collect(Collectors.toMap(BaseCaseFlag::getId, Function.identity()));
+
+        List<CasePartyFlagEntity> mergedCasePartyFlags = mergeFlagDetails(
+            existingPartyFlagEntityMap, incomingPartyFlags, pcsCaseEntity,
```
benouaer: you can just pass null instead pcsCaseEntity as it's not needed here

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -0,0 +1,176 @@
...
+                    && !incomingParty.getDefendantFlags().getDetails().isEmpty()) {
+                    mergePartyFlagGroup(incomingParty.getDefendantFlags(), partyEntity.getDefendantFlags(),
+                                        partyEntity, pcsCaseEntity, flow
+                    );
+                }
+            }
+        }
+    }
+
+    private void mergePartyFlagGroup(Flags incomingPartyFlags, List<CasePartyFlagEntity> existingFlags,
```
benouaer: no need for the pcsCaseEntity argument here

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -0,0 +1,176 @@
...
+
+            for (ListValue<Party> incomingPartyValue : incomingParties) {
+                Party incomingParty = incomingPartyValue.getValue();
+
+                PartyEntity partyEntity = existingPartiesMap.get(UUID.fromString(incomingPartyValue.getId()));
+
+                if (incomingParty.getDefendantFlags() != null
+                    && !incomingParty.getDefendantFlags().getDetails().isEmpty()) {
+                    mergePartyFlagGroup(incomingParty.getDefendantFlags(), partyEntity.getDefendantFlags(),
+                                        partyEntity, pcsCaseEntity, flow
```
benouaer: no need for the pcsCaseEntity argument here

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
aschalew-at: Latest push of the code removed the if check.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
aschalew-at: partyEntity is no more part of the if check.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
aschalew-at: latest code renamed existingCaseFlagEntitiesMap to existingFlagEntitiesMap. I will rename incomingCaseFlags to incomingFlags

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
aschalew-at: Done in latest code

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseFlagsView.java
```diff
@@ -0,0 +1,148 @@
...
+                .partyName(partyEntity.getOrgName() == null || partyEntity.getOrgName().isEmpty()
+                               ? partyEntity.getFirstName() + " " + partyEntity.getLastName()
+                               : partyEntity.getOrgName())
+                .roleOnCase(partyEntity.getOrgName() == null || partyEntity.getOrgName().isEmpty()
+                                ? DEFENDANT : CLAIMANT)
+                .details(new ArrayList<>())
+                .build();
+        }
+
+        List<BaseCaseFlag> respondentFlags = new ArrayList<>(partyEntity.getDefendantFlags());
```
benouaer: should this be called defendantFlags?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseFlagsView.java
```diff
@@ -0,0 +1,148 @@
...
+    private ListValue<Party> mapPartyWithDefendantFlags(PartyEntity partyEntity) {
+        return ListValue.<Party>builder()
+            .id(partyEntity.getId().toString())
+            .value(
+                Party.builder()
+                    .nameKnown(partyEntity.getNameKnown())
+                    .addressKnown(partyEntity.getAddressKnown())
+                    .phoneNumberProvided(partyEntity.getPhoneNumberProvided())
+                    .emailAddress(partyEntity.getEmailAddress())
+                    .firstName(partyEntity.getOrgName() == null || partyEntity.getOrgName().isEmpty()
```
benouaer: this check was already done in the filter

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseFlagsView.java
```diff
@@ -0,0 +1,148 @@
...
+                .map(paths -> ListValue.<String>builder()
+                    .id(paths[0])
+                    .value(paths[1])
+                    .build())
+                .toList();
+    }
+
+    private void mapComplexPartyFlagFields(PCSCase pcsCase, PcsCaseEntity pcsCaseEntity) {
+        List<ListValue<Party>> mappedParties = pcsCaseEntity.getParties().stream()
+            .filter(partyEntity -> partyEntity.getOrgName() == null || partyEntity.getOrgName().isEmpty())
```
benouaer, line 93: do party flags only apply to individuals?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseFlagsView.java
```diff
@@ -0,0 +1,148 @@
...
+                                 ? partyEntity.getFirstName() : partyEntity.getOrgName())
+                    .lastName(partyEntity.getLastName())
+                    .defendantFlags(mapDefendantFlags(partyEntity))
+                    .build()
+            )
+            .build();
+    }
+
+    private Flags mapDefendantFlags(PartyEntity partyEntity) {
+        if (partyEntity.getDefendantFlags() == null || partyEntity.getDefendantFlags().isEmpty()) {
```
benouaer: you can user CollectionUtils.isEmpty(partyEntity.getDefendantFlags()) to do both checks at the same time

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseFlagsView.java
```diff
@@ -0,0 +1,148 @@
...
+            return Flags.builder()
+                .partyName(partyEntity.getOrgName() == null || partyEntity.getOrgName().isEmpty()
```
benouaer: you can use !StringUtils.hasText(partyEntity.getOrgName()) to do both checks at the same time

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseServiceTest.java
```diff
@@ -240,6 +255,172 @@ void shouldPatchCaseDataWithCaseLinks() {
...
+        caseFlagEntity.setDefaultStatus("Active");
+        caseFlagEntity.getFlagRefData().setFlagCode("CF0008");
+        caseFlagEntity.setFlagComment("Police arrest inactive");
+        caseFlagEntity.setDateTimeModified(LocalDateTime.now());
+
+        List<CaseFlagEntity> caseFlagEntities = new ArrayList<>();
+        caseFlagEntities.add(caseFlagEntity);
+
+        return caseFlagEntities;
+    }*/
```
benouaer: should this be removed?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
benouaer: the party entity is guaranteed here because the flag can only exist on an existing party

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
benouaer: having thought about it, I don't think we need this logic here because xui does not maintain the flagUpdateComment field
we should be saving these fields in the same way they would be saved in centralised CCD
if we look at this functionality on a service that doesn't use decentralised CCD, it would behave differently because the fields would be updated automatically by xui
and given that the caseFlag component isn't managed by us, we shouldn't be trying to work around their 'bug'
if xui decide to fix that 'bug' at some point, then we would have to change our code
so we should just always persist both flagComment and flagUpdateComment regardless of status or flow

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseFlagsView.java
```diff
@@ -0,0 +1,144 @@
...
+                   .nameCy(caseFlagEntity.getFlagRefData().getFlagNameWelsh())
+                   .flagComment(caseFlagEntity.getFlagComment())
+                   .flagCommentCy(caseFlagEntity.getFlagCommentWelsh())
+                   .status(caseFlagEntity.getDefaultStatus())
+                   .subTypeKey(caseFlagEntity.getSubTypeKey())
+                   .subTypeValue(caseFlagEntity.getSubTypeValue())
+                   .subTypeValueCy(caseFlagEntity.getSubTypeValueWelsh())
+                   .flagUpdateComment(caseFlagEntity.getFlagUpdateComment() != null
+                                      ? caseFlagEntity.getFlagUpdateComment()
+                                      : caseFlagEntity.getFlagUpdateCommentWelsh())
```
benouaer: I don't think we should workaround xui bugs, we should just reload the data as we received it from xui

benouaer (approved):

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagServiceTest.java
```diff
@@ -0,0 +1,324 @@
...
+
+        // When
+        underTest.mergePartyFlags(incomingParties, pcsCaseEntity.getParties(), EventFlow.UPDATE.name());
+
+        // Then
+        assertEquals(1, pcsCaseEntity.getParties().size());
+        PartyEntity updatedParty = pcsCaseEntity.getParties().iterator().next();
+
+        assertNotNull(updatedParty.getDefendantFlags());
+        assertEquals(1, updatedParty.getDefendantFlags().size());
```
tvr-hmcts: assertThat

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -0,0 +1,171 @@
...
+            }
+
+            flagEntity.setParentEntity(pcsCaseEntity, partyEntity);
+
+            if ("CREATE".equals(flow)) {
+                flagEntity.setFlagComment(incomingFlagDetail.getFlagComment());
+                flagEntity.setFlagCommentWelsh(incomingFlagDetail.getFlagCommentCy());
+            } else {
+                if ("Inactive".equals(flagEntity.getDefaultStatus())) {
+                    flagEntity.setFlagUpdateComment(incomingFlagDetail.getFlagUpdateComment());
```
tvr-hmcts: The tests cover all routes apart from the setting of this.  Where the getDefaultStatus() is set as 'Inactive'

I mention as there is a getStatus() also which the tests do cover.

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagServiceTest.java
```diff
@@ -0,0 +1,324 @@
...
+    private List<ListValue<String>> createPathListValue(boolean isPathEmpty) {
+        List<ListValue<String>> paths = new ArrayList<>();
+
+        ListValue<String> path = ListValue.<String>builder()
+            .id(UUID.randomUUID().toString())
+            .value("Case")
+            .build();
+        paths.add(path);
+
+        return isPathEmpty ? null : paths;
```
tvr-hmcts: isPathEmpty is always passed in as false

tvr-hmcts (commented):
Just a few things spotted.

toby-plunkett (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -0,0 +1,147 @@
...
+
+            if (incomingParty.getDefendantFlags() != null
+                && !incomingParty.getDefendantFlags().getDetails().isEmpty()) {
+                mergePartyFlagGroup(incomingParty.getDefendantFlags(), partyEntity.getDefendantFlags(),
+                                    partyEntity);
+            }
+        }
+    }
+
+    private void mergePartyFlagGroup(Flags incomingPartyFlags, List<CasePartyFlagEntity> existingFlags,
```
benouaer: the existing flags variable isn't used/needed

benouaer: I think you can probably merge the mergePartyFlags and mergePartyFlagGroup methods now

aschalew-at: Refactored two methods into one

benouaer (approved):

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagServiceTest.java
```diff
@@ -0,0 +1,321 @@
...
+                       .hearingRelevant(YesOrNo.YES)
+                       .path(createPathListValue())
+                       .build())
+            .build();
+        flagDetails.add(flagDetailListValue);
+
+        return flagDetails;
+    }
+
+    private List<CasePartyFlagEntity> castToCasePartyFlag(List<BaseCaseFlag> caseFlagEntity) {
```
tvr-hmcts: Not used any more.

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagServiceTest.java
```diff
@@ -0,0 +1,321 @@
...
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.UUID;
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
tvr-hmcts: Minor: There is a mixture of these but I think the preference is to use assertj

tvr-hmcts (approved):
Couple of minor things

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseFlagsView.java
```diff
@@ -0,0 +1,138 @@
...
+                    .defendantFlags(mapDefendantFlags(partyEntity))
+                    .build()
+            )
+            .build();
+    }
+
+    private Flags mapDefendantFlags(PartyEntity partyEntity) {
+        if (CollectionUtils.isEmpty(partyEntity.getDefendantFlags())) {
+            return Flags.builder()
+                .partyName((requireNonNullElse(partyEntity.getFirstName(), "Person")
```
toby-plunkett, line 120: Is there anyone we can confirm the defaults with?

aschalew-at: We had a meeting with business, and business confirmed  that instead of empty party name, we can use 'Person Unknown'.

PR #1645, opened 2026-04-10
HDPI-5731: update notice details screen

gmmagruder:
### Jira link

See [HDPI-5731](https://tools.hmcts.net/jira/browse/HDPI-5731)

### Change description

- Updated content for email and other options in notice details page
- Renamed noticeEmailExplanation to noticeEmailAddress
- Changed input for noticeEmailAddress from multi line textbox to single line textbox and reduced character limit to 60

### Testing done

Manual testing locally

  * [ ] Yes
  * [x] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/NoticeServedDetails.java
```diff
@@ -80,15 +80,13 @@ public class NoticeServedDetails {
     private String noticePersonName;
 
     @CCD(
-            label = "Explain how it was served by email",
-            hint = "You can enter up to 250 characters",
-            typeOverride = TextArea
+            label = "What email address was the document sent to?",
+            hint = "For example, name@example.com"
```
scottstewart-sl: Maybe add `typeOverride = FieldType.Email` to get some validation of the input value, (if that is what you want).

Also since this is now a single line text field, you should hopefully just be able to add the `max` attribute on here and ExUI/CCD will do the length validation for you and you wouldn't need the custom validation in `NoticeDetailsService.java` . (Give that a quick test though to make sure it works with the Email type, if you do use that).

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/NoticeDetailsServiceTest.java
```diff
@@ -378,19 +378,19 @@ void shouldValidateEmailWithValidExplanation() {
...
         @Test
-        void shouldValidateEmailWithTooLongExplanation() {
+        void shouldValidateEmailWithTooLongEmailAddress() {
             // Given
             LocalDateTime pastDateTime = LocalDateTime.now().minusDays(1);
 
-            // Create a string that exceeds 250 characters
-            String longText = "0123456789".repeat(26); // 10 chars x 26 = 260 chars
+            // Create a string that exceeds 60 characters
+            String longText = "0123456789".repeat(7); // 10 chars x 7 = 70 chars
```
scottstewart-sl: I think 61 would be a better test since it is the boundary condition.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/NoticeServedDetails.java
```diff
@@ -80,15 +80,13 @@ public class NoticeServedDetails {
...
     )
-    private String noticeEmailExplanation;
+    private String noticeEmailAddress;
```
guygrewal77, line 89: Could this name change cause an issue with elastic search in the master build ? Or can an elastic search reindex be carried out somehow ? Might be worth checking with Praveen on this.

scottstewart-sl: Hopefully a name change would be OK, as long as it is not a name that has already been used with a different type. I think we can try this one with crossed fingers, rather than going to the effort of deprecating the old field name, deploying, and then removing the old field and deploying again.

If there is an issue with the build pipeline that mixes the staging and AAT envs then we should try to fix that instead of working around it.

guygrewal77 (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/NoticeServedDetails.java
gmmagruder: Updated to use typeOverride = FieldType.Email. Tested locally with max = 60 and ExUI validation was triggered.

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/NoticeDetailsServiceTest.java
gmmagruder: No longer needed as validation is done on the ExUI side.

scottstewart-sl (approved):

PR #1649, opened 2026-04-11
HDPI-5164: Add full replace draft save for respond to possession claim

arun-hmcts:
### Jira link

https://tools.hmcts.net/jira/browse/HDPI-5932

### Change description

Replace the defendant case data as a whole no patched saved .

### Testing done

Locally tested

  * [ ] Yes
  * [ ] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/respondpossessionclaim/SubmitEventHandler.java
```diff
@@ -27,6 +28,7 @@ public class SubmitEventHandler implements Submit<PCSCase, State> {
     private final ClaimResponseService claimResponseService;
     private final DefendantResponseService defendantResponseService;
 
+    @Transactional
```
scottstewart-sl, line 31: Minor - there's no harm in having this as far as I know, but I think these cals are already in a transaction by the time we get here.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/respondpossessionclaim/ClaimResponseService.java
```diff
@@ -34,9 +34,6 @@ public class ClaimResponseService {
     private final SecurityContextService securityContextService;
     private final ModelMapper modelMapper;
 
-    //when contact by phone = no, we MUST skip saving contact_by_text and phone number.
-    private boolean shouldSavePhoneNumAndTextPreference;
```
scottstewart-sl: Looks like you've fixed an undiscovered bug there!

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftCaseDataService.java
```diff
@@ -78,6 +78,30 @@ public boolean hasUnsubmittedCaseData(long caseReference, EventId eventId) {
...
+        UUID userId = getCurrentUserId();
+        log.info("Saving draft: caseReference={}, eventId={}, userId={}", caseReference, eventId, userId);
+
+        String eventDataJson = writeCaseDataJson(eventData);
+
+        DraftCaseDataEntity draftCaseDataEntity = draftCaseDataRepository
+            .findByCaseReferenceAndEventIdAndIdamUserId(caseReference, eventId, userId)
+            .orElseThrow(() -> new UnsubmittedDataException(
+                "No draft found for caseReference=" + caseReference
+                    + ", eventId=" + eventId + ", userId=" + userId));
```
scottstewart-sl, line 95: I wonder if we even need this, (and the call in the start handler that sets up an empty draft). Could we just make this call create the entity if it is missing? We have the full draft JSON to store anyway.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/respondpossessionclaim/ClaimResponseService.java
```diff
@@ -150,5 +150,6 @@ private void saveContactPreferences(PartyEntity party, DefendantResponses defend
         }
```
scottstewart-sl: Not your change, but I suspect this isn't needed as `party` is already a managed entity that was fetched from the DB

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/respondpossessionclaim/SubmitEventHandler.java
arun-hmcts: Scott — my understanding is that the CCD callback is only an HTTP call into our API. It does not provide a DB transaction boundary for the writes we do afterwards. Since this method performs multiple related database operations, @Transactional is needed on our side so they are treated atomically. Otherwise we risk partial saves if one of the later calls fails.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftCaseDataService.java
arun-hmcts: The START callback creates the draft, so by the time save is called the draft should always exist. The orElseThrow is intentional — if it's missing at that point, something has gone wrong.

scottstewart-sl: Yes I was suggesting that maybe we don't create the empty draft in the start callback. The `caseData` only going to get overridden in mid-event callbacks so we can recreate it for any request. So even if the draft data was missing for some reason by the time the user got to, say, page 6 then we would still have all the answers for all the previous pages in the payload wouldn't we?

It's not a problem like you have it anyway.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/party/PartyEntity.java
```diff
@@ -82,10 +84,14 @@ public class PartyEntity {
     @OneToOne(cascade = ALL)
     private ContactPreferencesEntity contactPreferences;
 
+    /** Set by claimant at case creation. Must not be modified by defendant responses. */
+    @Column(updatable = false)
```
scottstewart-sl: I'm not sure about this. Although I agree that the defendant responses shouldn't change this, the defendant may provide their correspondence address anyway and a casework could update the case data with that address once it has been accepted, which means that this column would need to be updatable.

It's only in `SubmitEventHandler` where the draft data for the Make a Claim response is written to the relational tables isn't it? So as long as that code doesn't try to update any party entities it should be fine to not have these annotations here.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/respondpossessionclaim/PaymentAgreement.java
```diff
@@ -26,4 +30,14 @@ public class PaymentAgreement {
...
 
+    @CCD
+    private VerticalYesNo repayArrearsInstalments;
+
+    @CCD(typeOverride = FieldType.MoneyGBP)
+    @JacksonMoneyGBP
+    private BigDecimal additionalRentContribution;
+
+    @CCD
+    private String additionalContributionFrequency;
```
scottstewart-sl, line 41: These changes seem unrelated to the main purpose of this PR.

arun-hmcts: Found during local testing — the DTO was missing fields the entity and V068 already have, so the holistic round-trip drops them on resume/submit; happy to revert if you'd rather have them in a separate PR.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/party/PartyEntity.java
arun-hmcts: Good point — agreed. Removed @Column(updatable = false) from nameKnown, addressKnown, and addressSameAsProperty so caseworker flows can update them later.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftCaseDataService.java
arun-hmcts: You're right — the full caseData is in every request payload, will need a completed testing again. will see I can do it as part of the pr today or will do as part of another refactoring ticket ..

scottstewart-sl (approved):

libanAbdirahman1 (approved):

PR #1727, opened 2026-04-23
HDPI-5201: Prevent duplicate gen app submissions

scottstewart-sl:
### Jira link

See [HDPI-5201](https://tools.hmcts.net/jira/browse/HDPI-5201)

### Change description

Prevent duplicate gen app submissions from the pcs-frontend by requiring a unique client reference with each submission. This is to catch the unlikely event of the payload being sent twice for some reason.

Related pcs-frontend PR: https://github.com/hmcts/pcs-frontend/pull/1198

### Testing done

Tested locally with the corresponding changes in pcs-frontend.

  * [ ] Yes
  * [x] No

### Checklist


adusumillipraveen (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/citizen/CitizenCreateGenApp.java
```diff
@@ -51,7 +55,18 @@ private SubmitResponse<State> submit(EventPayload<PCSCase, State> eventPayload)
...
-        genAppService.createGenAppEntity(caseData, pcsCaseEntity, applicantParty);
+        CitizenGenAppRequest citizenCreateGenApp = caseData.getCitizenGenAppRequest();
+
+        String clientReference = citizenCreateGenApp.getClientReference();
+        if (clientReference == null) {
+            return errorResponse("No client reference in request");
+        }
+
+        if (genAppRepository.existsByClientReference(clientReference)) {
+            return errorResponse("Application alreadys exists for client reference");
```
tvr-hmcts: Typo in 'already'

tvr-hmcts (commented):
One minor and then can all the tests within the GenAppServiceTest also assert that there is a call to the pcsCaseEntity.addGenApp(

As I suspect that is important but all the tests pass if I remove it - suggesting it is not.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/citizen/CitizenCreateGenApp.java
scottstewart-sl: Now fixed

scottstewart-sl:
> One minor and then can all the tests within the GenAppServiceTest also assert that there is a call to the pcsCaseEntity.addGenApp(
> 
> As I suspect that is important but all the tests pass if I remove it - suggesting it is not.

@tvr-hmcts I've added a test to specifically test that the addGenApp call is made with the expected value and confirmed that it fails with the line commented out. I don't think it should be in every test, in the same way that, for example, we don't check that the languageUsed field is set correctly in every test.

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/genapp/GenAppServiceTest.java
```diff
@@ -71,21 +70,36 @@ void shouldReturnSavedEntity() {
...
+            .build();
+
+        GenAppEntity savedGenAppEntity = mock(GenAppEntity.class);
+        when(genAppRepository.save(isA(GenAppEntity.class))).thenReturn(savedGenAppEntity);
+
+        // When
+        underTest.createGenAppEntity(genAppRequest, pcsCaseEntity, applicantParty);
+
+        // Then
+        verify(pcsCaseEntity).addGenApp(genAppEntityCaptor.capture());
```
tvr-hmcts, line 99: Cool ok ta

tvr-hmcts:
Yep fair enough.  So long as there is now one test asserting that it happens then that should be good enough.  :)

tvr-hmcts (approved):

PR #1752, opened 2026-04-29
HDPI-5499-Service-callback-completion

tvr-hmcts:
https://tools.hmcts.net/jira/browse/HDPI-5499

This PR associates a fee payment with a specific claimant party (the responsible party) when creating a service request.

FeePaymentEntity — Added a @ManyToOne relationship to PartyEntity, linking each fee payment to the party responsible for it via a new party_id column.

PaymentService — Updated saveNewFeePayment to accept a responsibleParty (organisation name) parameter. A new helper method retrieveClaimPartyEntity resolves the matching ClaimPartyEntity from the claim, throwing PartyNotFoundException if no match is found. The payment entity is now persisted with the resolved party reference.

ServiceRequestUpdate — Added a getPaymentReference() convenience method to safely retrieve the payment reference, guarding against a null payment object.

Integration tests and Unit tests


src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentService.java
```diff
@@ -124,4 +130,12 @@ private ClaimEntity retrieveClaimEntity(Long caseReference) {
         return pcsCaseEntity.getClaims().getFirst();
     }
 
+    private ClaimPartyEntity retrieveClaimPartyEntity(ClaimEntity claimEntity, String responsibleParty) {
+        return claimEntity.getClaimParties()
+            .stream()
+            .filter(party -> responsibleParty.equals(party.getParty().getOrgName()))
```
scottstewart-sl, line 136: Could we use the responsible party ID for this instead? It's a more reliable way to find a party and would also mean that (a) we're not relying on the payment responsible org name being exactly the same as the persisted name and (b) it would work with individuals as well

tvr-hmcts: That would make sense.  I followed the pattern already take from within the MAC submission TBH

The Task is loose as it is just a String (which Im not a fan of) so it is feasible and if needed for individuals then I'll do that yep.

I will test up here first and then take a look.

tvr-hmcts: Following on from the Slack this appears to be a wider concern.  It will be an issue within the callback handling also to identify the source of the payment from the callback so that a relevant 'CallBackHandler' implementation can be applied for each journey.

Also that the claimant id can not be derived at the time so to make the submission use that within the fee scheduling.

A ticket has been raised to address this and the other usage patterns needed from the callback processing:

https://tools.hmcts.net/jira/browse/HDPI-5244

toby-plunkett (approved):

scottstewart-sl (approved):

libanAbdirahman1 (approved):

PR #1779, opened 2026-05-05
Add suppport for view documents screen

adusumillipraveen:
### Jira link

https://tools.hmcts.net/jira/browse/HDPI-5549 

### Change description

- uses alldocuments to return list of documents to show in CUI, idea is to filter them in future if a certain user shouldn't have access to certain documents.
- adds submitted time for docs into document table
- Moves documents to a different task group

### Testing done

Tested with https://github.com/hmcts/pcs-frontend/pull/1276 

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/PCSCaseView.java
```diff
@@ -253,6 +253,10 @@ private List<ListValue<Document>> mapAndWrapDocuments(PcsCaseEntity pcsCaseEntit
                            .url(entity.getUrl())
                            .binaryUrl(entity.getBinaryUrl())
                            .categoryId(entity.getCategoryId())
+                           .uploadTimestamp(entity.getSubmittedDate() == null
```
adusumillipraveen, line 226: Reusing uploadTimestamp to avoid creating a full custom bean as business wants to show the date the draft is submitted , not essentially when its uploaded

src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/DocumentEntity.java
```diff
@@ -71,4 +77,11 @@ public class DocumentEntity {
     @ManyToOne(fetch = LAZY)
     @JoinColumn(name = "counter_claim_id")
     private CounterClaimEntity counterClaim;
+
+    @PrePersist
+    void prePersist() {
+        if (submittedDate == null) {
+            submittedDate = Instant.now();
```
scottstewart-sl, line 108: Minor - have you checked that this timestamp is persisted as UTC even when the JVM is running as UK time? I think it's probably fine when you use `Instant`, but worth checking. Looking at the code we have a mix of Instant and LocalDateTime in the entities, with the latter being explicitly set from the UTC `Clock` instance, (or should be at least).

adusumillipraveen: good shout, may be we should set  ?

```
hibernate:
  properties
    jdbc:
      time_zone: UTC
```

scottstewart-sl: I think we want the timestamps in the JVM to also be in UTC don't we, unless it for checking the current day when validating a field, (e.g. date must be before today). `Instant` is probably OK as long as it gets serialised to JSON in UTC, (which is what the pcs-frontend would expect for example), as is `LocalDateTime` when it was derived from a UTC `Clock`

scottstewart-sl (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/DocumentEntity.java
adusumillipraveen: true, i think we need to align the repo a bit across. On this one, i checked locally and it saves in UTC.

fudge88 (approved):

aqibh25 (approved):

PR #1818, opened 2026-05-13
HDPI-5822: Notification - Claim Saved For Later

lucas-phillips28:
### Jira link

See [HDPI-5822](https://tools.hmcts.net/jira/browse/HDPI-5822)

### Change description
- Send email notification when claimant saves claim for later
- Refactor personalisation into more readable objects

### Testing done
Tested locally

  * [ ] Yes
  * [x] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/notify/model/NotificationType.java
```diff
@@ -0,0 +1,24 @@
...
+    @Override
+    public String toString() {
+        return value;
+    }
+
+    public static NotificationType fromString(String status) {
+        return Arrays.stream(values())
+            .filter(s -> s.value.equalsIgnoreCase(status))
+            .findFirst()
+            .orElseThrow(() -> new IllegalArgumentException("Unknown status: " + status));
```
scottstewart-sl: Minor - this method mentions "status" in a few places, perhaps left over from a copy and paste?

Btw, are the `toString()` and the `value` field of this enum ever used anywhere?

lucas-phillips28: Resolved- I've just removed this completely as this code is unneeded

scottstewart-sl (approved):

src/main/java/uk/gov/hmcts/reform/pcs/notify/template/personalisation/TemplatePersonalisation.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.hmcts.reform.pcs.notify.template.personalisation;
+
+import java.util.Map;
+
+public interface TemplatePersonalisation {
```
tvr-hmcts, line 5: Could a Record work?

tvr-hmcts (approved):
LGTM - just one comment to consider.

src/main/java/uk/gov/hmcts/reform/pcs/notify/template/personalisation/TemplatePersonalisation.java
lucas-phillips28: I had considered this, but I felt like the builder was more readable than the record- but I'm happy to switch to records if we prefer that

adusumillipraveen (commented):
Has this approach been discussed with the Tech Leads? I have some concerns about using in-memory event listeners here, based on experience from other HMCTS projects. If the service is deployed or the pod restarts after the transaction commits but before the listener completes, the event can be lost without any audit trail. My preference would be to persist the notification work as part of the same transaction, for example via db-scheduler or an outbox-style table, so we have a durable record, auditability, and DB-backed retry semantics.

adusumillipraveen (changes requested):

src/main/resources/db/migration/V109__make_case_notification_ids_nullable.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE case_notification ALTER COLUMN case_id DROP NOT NULL;
```
linusnorton: is this okay? seems like we're potentially losing some valuable constraints

linusnorton:
Also, see Praveen's note about in memory... the `notificationService.sendClaimantDraftSavedForLater(caseReference, pcsCase);` should be inside the transaction.

src/main/resources/db/migration/V109__make_case_notification_ids_nullable.sql
lucas-phillips28: I've removed this specific line as I have found a way to get the associated case. However, for the rest of the ids, they must be nullable because at the point where a notification is sent in this pr (whilst the claim is in draft), the party and claim have not been created in the db

lucas-phillips28:
> Has this approach been discussed with the Tech Leads? I have some concerns about using in-memory event listeners here, based on experience from other HMCTS projects. If the service is deployed or the pod restarts after the transaction commits but before the listener completes, the event can be lost without any audit trail. My preference would be to persist the notification work as part of the same transaction, for example via db-scheduler or an outbox-style table, so we have a durable record, auditability, and DB-backed retry semantics.

I've refactored the event listeners system to to use db-scheduler as requested

src/main/resources/application.yaml
```diff
@@ -237,6 +238,16 @@ role-assignment:
     max-retries: ${ROLE_ASSIGNMENT_MAX_RETRIES:5}
     backoff-delay-seconds: ${ROLE_ASSIGNMENT_BACKOFF_DELAY_SECONDS:3600s}
 
+counter-claim-notification:
+  request:
+    max-retries: ${COUNTER_CLAIM_NOTIFICAITON_MAX_RETRIES:5}
```
adusumillipraveen: Minor, but there is a typo in the env var name

adusumillipraveen (approved):

src/main/resources/application.yaml
lucas-phillips28: Resolved- also fixed the misspelling for the backoff delay env var as well

libanAbdirahman1 (approved):

PR #1825, opened 2026-05-13
HDPI-5244-Update-the-case-state

tvr-hmcts:
[HDPI-5244](https://tools.hmcts.net/jira/browse/HDPI-5244)

Case moves from PENDING_CASE_ISSUED to CASE_ISSUED state.

Enabling the case state change for the claim journey.   With accommodation for other payment approaches to plug into the payment callback.


src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/ResumePossessionClaim.java
```diff
@@ -354,24 +355,21 @@ private ClaimantInformation getClaimantInfo(PCSCase caseData) {
...
+            .journeyId(JourneyId.RESUME_POSSESSION_CLAIM)
             .build();
 
         schedulerClient.scheduleIfNotExists(
             FEE_CASE_ISSUED_TASK_DESCRIPTOR
-                .instance(taskId)
+                .instance(UUID.randomUUID().toString())
                 .data(taskData)
-                .scheduledTo(Instant.now())
+                .scheduledTo(Instant.now().plusSeconds(10))
```
toby-plunkett: Worth being configurable?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/service/CcdUpdateService.java
```diff
@@ -0,0 +1,40 @@
...
+import uk.gov.hmcts.reform.ccd.client.model.StartEventResponse;
+import uk.gov.hmcts.reform.pcs.ccd.domain.PCSCase;
+import uk.gov.hmcts.reform.pcs.idam.IdamService;
+
+import static uk.gov.hmcts.reform.pcs.ccd.event.EventId.payment;
+
+@AllArgsConstructor
+@Service
+@Slf4j
+public class CcdUpdateService {
```
toby-plunkett: I feel like this Service name is a little generic? Open to suggestions but perhaps CcdStateUpdateService.

toby-plunkett: On another hand, it's calling events specifically, so maybe CcdPaymentStateUpdateService? Up to you

toby-plunkett (commented):
a couple of minor things

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/service/CcdUpdateService.java
tvr-hmcts: Initial intention was to make it generic but back tracked from for now.  Made the name change as suggested.

src/integrationTest/java/uk/gov/hmcts/reform/pcs/controllers/PaymentCallBackControllerIT.java
```diff
@@ -57,18 +65,33 @@ public class PaymentCallBackControllerIT extends AbstractPostgresContainerIT {
...
     private FeeDto feeDto;
+    private FeesAndPayTaskData feesAndPayTaskData;
+    private ClaimEntity claimEntity;
 
     @BeforeEach
-    void setUp() {
+    void setUp() throws JsonProcessingException {
         serviceCaseReference = UUID.randomUUID().toString();
+        feesAndPayTaskData = Instancio.create(FeesAndPayTaskData.class);
+        feesAndPayTaskData.setCaseReference("1234");
```
scottstewart-sl: Might it be simpler to keep the case reference as a constant rather than setting it with a literal here and then getting it again to set the field, (which is then only used by establishTestCase)?

Also in my opinion it is simpler to keep it as a `long`, as that's how we represent it in most of the rest of the code and the conversion from long to string is then more reliable than going from a string to long.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/service/CcdPaymentStateUpdateService.java
```diff
@@ -0,0 +1,50 @@
...
+        log.debug("Submitting payment event for case: {}", caseId);
+        StartEventResponse startEventResponse = coreCaseDataApi.startEvent(idamToken, serviceAuthorization,
+                                                                           caseId, payment.name());
+        log.debug("StartEventResponse: {}", startEventResponse);
+        CaseDataContent submitContent = CaseDataContent.builder().event(Event.builder().id(payment.name()).build())
+            .eventToken(startEventResponse.getToken()).data(toJsonNode(PCSCase.builder().build())).build();
+        log.debug("submitContent: {}", submitContent);
+        CaseResource caseResource = coreCaseDataApi.createEvent(idamToken, serviceAuthorization,
+                                                                caseId, submitContent);
+        log.debug("CaseResouce response : {}", caseResource);
```
scottstewart-sl: Minor - it's only a debug log but there is a typo in CaseResource.

Do you actually need to keep these debug log lines now? (No problem if you do).

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentCallbackStrategyFactory.java
```diff
@@ -0,0 +1,21 @@
...
+@AllArgsConstructor
+public class PaymentCallbackStrategyFactory {
+
+    private final MakeAClaimPaymentCallbackStrategy makeAClaimPaymentCallbackStrategy;
+
+    public PaymentCallbackStrategy getStrategy(JourneyId journeyId) {
+        return Map.of(
+            JourneyId.RESUME_POSSESSION_CLAIM, makeAClaimPaymentCallbackStrategy
+        ).get(journeyId);
+    }
```
scottstewart-sl, line 23: I think it might be clearer if these class names used the term `PaymentCallbackHandler`. That way it is clear that it is the class that actually handles the callback, not just provides a strategy.

I also feel like the `JourneyId` is sitting half way between something that identifies the event that invoked the payment and the handler that should be used to handle the payment callback. As I see it, there is (currently) no need to know what event invoked the payment when handling the callback, so this enum might be better named as something that is intended to specifically identify the callback handler, (i.e. a 1:1 mapping). If it was all in code we might just register the method reference, but since it has to be persisted as a scheduled task, we need something that can be serialised.

So how about renaming `JourneyId` to `CallbackHandlerId` or something like that, (and losing the connection to `EventId`)?

scottstewart-sl (approved):

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/event/PaymentEvent.java
```diff
@@ -24,15 +24,14 @@ public class PaymentEvent implements CCDConfig<PCSCase, State, UserRole> {
     public void configureDecentralised(DecentralisedConfigBuilder<PCSCase, State, UserRole> configBuilder) {
         configBuilder
             .decentralisedEvent(payment.name(), this::submit)
-            .forStates(State.PENDING_CASE_ISSUED)
+            .forStates(State.PENDING_CASE_ISSUED, State.CASE_ISSUED)
```
toby-plunkett, line 27: If state = Case Issued, hasn't the user already paid?

tvr-hmcts: PENDING_CASE_ISSUED being from the Resume journey.   We are setting the state to CASE_ISSUED here in the event of a successful payment ... which curiously now Ive typed that ... means I need to check for that.

toby-plunkett (approved):

PR #1847, opened 2026-05-19
HDPI-5535 Page - View the response

mazzopardi2:
### Jira link

See [HDPI-5535](https://tools.hmcts.net/jira/browse/HDPI-5535)
See [HDPI-6457] (https://tools.hmcts.net/jira/browse/HDPI-6457)

### Change description

- New “view defendant response” flow - Defendants can load a submitted response via CCD event 
- Read mapping - Submitted response data is mapped from Postgres into the case payload the frontend uses.
- date submitted and statement-of-truth details are stored on submit.
- Claim issued date - New claim_issued_date on the claim; set when the issue fee is paid.
- Claimant details on view - Claimant name and address for service are added to the view payload from case parties.

### Testing done
-added unit tests and manual testing

  * [ ] Yes
  * [x] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/respondpossessionclaim/DefendantResponses.java
```diff
@@ -116,6 +117,13 @@ public class DefendantResponses {
...
     @CCD(max = 6400)
     private String otherConsiderationsDetails;
+    
     @CCD
     private VerticalYesNo makeCounterClaim;
+
+    @CCD
+    private LocalDateTime responseSubmittedDate;
+
+    @CCD(access = {CitizenAccess.class})
```
scottstewart-sl: Minor - does this actually need the Citizen access or does it inherit it from where the `DefendantResponses` field is added in `PossessionClaimResponse`?

Also can it have a max length annotation attribute?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/DefendantResponseView.java
```diff
@@ -0,0 +1,78 @@
...
+            Optional.ofNullable(claimant.getAddress())
+                .map(addressMapper::toAddressUK)
+                .ifPresent(response::setClaimantServiceAddress);
+        });
+
+        caseData.setPossessionClaimResponse(response);
+        return caseData;
+    }
+
+    private static Optional<PartyEntity> findPrimaryClaimant(PcsCaseEntity caseEntity) {
```
scottstewart-sl, line 81: There's a method to do more or less this being added to `PartyService` in a different PR:

https://github.com/hmcts/pcs-api/pull/1856/changes#diff-a65522037bc61bc6a19dc735999ef32417cafc4bf5693adfe26cd3bb73e62caa

You could either copy those changes into your branch or make a note to come back and refactor in the future, just to remove the duplication.

Unless you need to handle a scenario where the claimant party hasn't been created yet?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/respondpossessionclaim/DefendantResponseReadMapper.java
```diff
@@ -0,0 +1,305 @@
...
+            .isClaimAmountKnown(entity.getIsClaimAmountKnown())
+            .claimAmount(entity.getClaimAmount())
+            .estimatedMaxClaimAmount(entity.getEstimatedMaxClaimAmount())
+            .needHelpWithFees(entity.getNeedHelpWithFees())
+            .counterClaimFor(entity.getCounterClaimFor())
+            .counterClaimReasons(entity.getCounterClaimReasons())
+            .build();
+    }
+
+    private static YesOrNo toYesOrNo(VerticalYesNo value) {
```
scottstewart-sl: Minor - the `YesOrNoConverter` already has one of these methods that you can use

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/respondpossessionclaim/DefendantResponseService.java
```diff
@@ -206,6 +221,44 @@ private void buildAndLinkChildEntities(
...
+    ) {
+        if (StringUtils.isBlank(responses.getStatementOfTruthCompletedBy())) {
+            return;
+        }
+        StatementOfTruthCompletedBy completedBy = parseStatementOfTruthCompletedBy(
+            responses.getStatementOfTruthCompletedBy()
+        );
+        String fullName = Stream.of(party.getFirstName(), party.getLastName())
+            .filter(StringUtils::isNotBlank)
+            .collect(Collectors.joining(" "));
```
scottstewart-sl, line 297: Not for now, but we have a few places where we build the name for a PartyEntity, some of them have the blank check like yours and others not. In the future we could refactor this logic out into a method in, say, `PartyService`

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/respondpossessionclaim/DefendantResponseService.java
```diff
@@ -206,6 +221,44 @@ private void buildAndLinkChildEntities(
...
+        if (fullName.isBlank()) {
+            fullName = StatementOfTruthCompletedBy.DEFENDANT == completedBy
+                ? "Defendant"
+                : completedBy.name();
+        }
```
scottstewart-sl, line 300: In other journeys, (including Citizen gen apps), the statement of truth page explicitly asks the user for their full name as part of that page. I wonder if this should be doing the same thing rather than using their party name. (There might be some significance in the user typing out their full name for example, instead of signing something).

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/respondpossessionclaim/DefendantResponseService.java
```diff
@@ -206,6 +221,44 @@ private void buildAndLinkChildEntities(
...
+    
+        defendantResponse.setStatementOfTruth(
+            StatementOfTruthEntity.builder()
+                .completedBy(completedBy)
+                .accepted(YesOrNo.YES)
```
scottstewart-sl, line 305: I don't think this is an issue, but it would be nice if this was the value of the checkbox that the user would have filled in, even if it is always YES by the time the code gets here. If there was a bug in the front end  for example that meant that the user could submit the SoT page without ticking the box, then it would be better to at least record that (unticked) answer.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/respondpossessionclaim/DefendantResponses.java
mazzopardi2: It should inherit from the parent. Dropped it - great point!
removed redundant CitizenAccess since it inherits from parent.
Added max 100chars to statementOfTruthCompletedBy

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/DefendantResponseView.java
mazzopardi2: Good spot, yep this mirrors the same function in PR1856. 
For a submitted response on an issued claim, claimant should always exist, so I'm happy to refactor to PartyService once 1856 is merged. Added a TODO so I can come back to it.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/respondpossessionclaim/DefendantResponseReadMapper.java
mazzopardi2: yep replaced good point!

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/respondpossessionclaim/DefendantResponseService.java
mazzopardi2: Agreed the name building is inconsistent across the codebase. Added a TODO to come back to this

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/respondpossessionclaim/DefendantResponseService.java
mazzopardi2: Agreed, typed full name on SoT is the right pattern (like gen apps). At the moment currently derive that on submit from party data because respond SoT doesn’t collect a name yet. I'll follow up on capture in the respond journey. Added a TODO on linkStatementOfTruth.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/respondpossessionclaim/DefendantResponseService.java
mazzopardi2: At the moment, I've hardcoded accepted to YES because we don't have a defendant SoT checkbox on the payload yet. I'll do the same as GenApps when we add SoT to respond submit, so we record the actual tick. Happy to leave as-is for this PR given it's more view-focused but I added a TODO.

scottstewart-sl (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/respondpossessionclaim/PossessionClaimResponse.java
```diff
@@ -35,6 +37,12 @@ public class PossessionClaimResponse {
     )
     private List<ListValue<String>> claimantOrganisations;
 
+    @CCD(access = {CitizenAccess.class}, label = "Name")
```
toby-plunkett: I thought the PossessionClaimResponse was submitted by the defendant? Wouldn't that make this the defendant name? Or am I misinterpreting this

mazzopardi2: No you're right, the PossessionClaimResponse is submitted by the defendant, but it also carries the claimant details that we show on the view-the-repsonse screen (along with the defendant submitted answers). So this is the label for the claimant field next to the claimant service address in the claimant details section

toby-plunkett: oh I see! That makes sense.

toby-plunkett (approved):

PR #1848, opened 2026-05-19
Hdpi 5794 update lr contact details

m014389b:
### Jira link

See [HDPI-5794](https://tools.hmcts.net/jira/browse/HDPI-5794)

### Change description

image

Update legal representative details. Only applies for linked defendant solicitors

DB:  
- added legal_representative_organisation table (to store a legal reps org that is linked to a defendant)
- added party_legal_rep_org table  (table to link legal_rep_org and parties)
- add organisation_uuid to legal_representative (to create a new link between legal_rep and legal_rep_org in the db whilst keeping legal_rep.organisation_id to be backwards compatible )
- added "legal_representative_organisation_id" to "draft_case_data" to enable retrieval of drafts for a specific party + legal rep org
- add new index "draft_case_data_unique_by_organisation_idx" (ensure uniqueness of case, event, legal rep org and party combination"

Modelling is now a case can have many unique legal representative organisations LRO (unique by organisation ID). Where a LRO represents multiple defendants on a case there will still be a single instance of a LRO.

Introduced a env for pcs-frontend-url to allow for redirecting via a hyperlink of the summary page. 

Once a LRO has updated their contact details then the above hyperlink text will be exposed and the "Respond to claim" event will be visible. The linking will redirect to the configured frontend url with the case loaded.

Updated references to legal representative to use legal representative organisation where relevant. It was decided that legal representative organisation won't store individual details such as first name etc

### Testing done

Manual testing and unit tests

  * [ ] Yes
  * [x] No

### Checklist


src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/legalrepresentative/LegalRepresentativePageServiceTest.java
```diff
@@ -0,0 +1,274 @@
...
+
+        // when
+        legalRepresentativeService.save(userIdamId, legalRepresentativeDetails);
+
+        // then
+        assertEquals(email, legalRepresentativeEntity.getEmail());
+    }
+
+    @Test
+    void save_WithUseEmailYes_DoesNotSetEmail() {
```
sadmanrahh: Could we add a test for updating an existing saved custom value back to the default/registered option, e.g. useEmailAddress == YES after a custom email was previously saved? That would cover stale values being cleared or reset.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/LegalRepresentativeDetails.java
```diff
@@ -0,0 +1,64 @@
...
+    )
+    private String reference;
+
+    @CCD(label = "Do you want to provide a contact phone number?")
+    private VerticalYesNo provideContactPhoneNumber;
+
+    @CCD(label = "Enter phone number",
+        regex = "^\\s*0\\d{10}\\s*$",
+        max = 11)
+    private String contactPhoneNumber;
```
sadmanrahh, line 42: The phone number regex doesn’t fully match AC07. It currently allows any 11-digit number starting with 0, so prefixes like 04/05/06 would pass, and valid 10-digit landlines starting 01/02 would fail. Could we tighten the regex and add tests for the AC07 examples?

m014389b: Done tightened up

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/legalrepresentative/LegalRepresentativeSummaryService.java
```diff
@@ -0,0 +1,86 @@
...
+        } else {
+            pcsCase.setSummaryLegalRepresentativeMarkdown(UPDATE_DETAILS_MARKDOWN
+                                                              .formatted(legalRepresentativeContactDetails));
+        }
+    }
+
+    private Optional<PartyLegalRepresentativeOrganisationEntity> isActivelyLinkedToAnyDefendant(PcsCaseEntity
+                                                                                                    pcsCaseEntity) {
+
+        return pcsCaseEntity.getParties().stream()
```
sadmanrahh: This checks all active party legal-rep org links, but does not confirm the party is a defendant. Since the ticket says this is only available for defendant LR, can we filter to defendant parties only or reuse the existing defendant LR validator?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/legalrepresentativedetails/LegalRepresentativeContactDetailsPage.java
```diff
@@ -0,0 +1,167 @@
...
+            .mandatoryWithLabel(AddressUK::getPostCode, "Postcode")
+            .done()
+            .label("legalRepresentativeDetails-phoneNumber-question", """
+                    ----
+                    <h2 class="govuk-heading-m">Contact phone number</h2>
+                    <p class="govuk-body-m ">
+                        You should provide a phone number so we can contact you if there are urgent updates.
+                    </p>
+                    """)
+                .optional(LegalRepresentativeDetails::getProvideContactPhoneNumber)
```
sadmanrahh, line 130: AC04 says the user must select a radio option for each question, but provideContactPhoneNumber is optional here. Can this be mandatory, while keeping contactPhoneNumber mandatory only when the answer is YES?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/legalrepresentative/LegalRepresentativePageService.java
```diff
@@ -0,0 +1,125 @@
...
+            legalRepresentativeOrganisationEntity.addLegalRepresentative(legalRepresentative);
+        }
+
+        if (legalRepresentativeDetails.getDifferentPostalAddress() != null
+            && legalRepresentativeDetails.getDifferentPostalAddress().equals(VerticalYesNo.YES)) {
+            legalRepresentativeOrganisationEntity.setAddress(mapAddressUkToAddressEntity(legalRepresentativeDetails
+                                                                .getUpdatedCorrespondenceAddress()));
+        }
+
+        if (legalRepresentativeDetails.getProvideContactPhoneNumber() != null
```
sadmanrahh, line 52: If the LR previously saved a phone number and later selects NO, the old phone number remains on the entity. Can we explicitly clear phone when provideContactPhoneNumber == NO?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/legalrepresentative/LegalRepresentativePageService.java
```diff
@@ -0,0 +1,125 @@
...
+            && legalRepresentativeDetails.getProvideContactPhoneNumber().equals(VerticalYesNo.YES)) {
+            legalRepresentativeOrganisationEntity.setPhone(legalRepresentativeDetails.getContactPhoneNumber());
+        }
+
+        if (legalRepresentativeDetails.getReference() != null && !legalRepresentativeDetails.getReference().isEmpty()) {
```
sadmanrahh, line 58: The reference is optional, but clearing it will not clear the persisted contactReference; the old value remains. Can we persist the submitted value, including blank/null?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/legalrepresentative/LegalRepresentativePageService.java
```diff
@@ -0,0 +1,125 @@
...
+            legalRepresentativeOrganisationEntity.setContactReference(legalRepresentativeDetails.getReference());
+        }
+
+        if (legalRepresentativeDetails.getUseEmailAddress() != null
```
sadmanrahh: If the LR previously entered a custom email and later chooses to use the registered email, the old custom email remains persisted. Can we clear/reset email when useEmailAddress == YES?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/legalrepresentativedetails/LegalRepresentativeContactDetailsPage.java
m014389b: The AC04 is referring to mandatory questions. Contact phone number is optional

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/legalrepresentative/LegalRepresentativePageService.java
m014389b: That is correct. The entity only updates if a value is provided

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/legalrepresentative/LegalRepresentativePageService.java
m014389b: That is correct. The entity only updates if a value is provided

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/legalrepresentative/LegalRepresentativePageService.java
m014389b: The organisation email is used at 1st, then if any custom email is used that takes priority

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/legalrepresentative/LegalRepresentativePageServiceTest.java
m014389b: As mentioned, "default" is no longer a valid scenario

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/legalrepresentative/LegalRepresentativeSummaryService.java
m014389b: The party legal rep link would only be created for defendants but makes sense to add an extra defensive layer

sadmanrahh (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/LegalRepresentativeDetails.java
```diff
@@ -0,0 +1,64 @@
...
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class LegalRepresentativeDetails {
+
+    @CCD(label = "Do you want to use this email address for notifications?")
+    private VerticalYesNo useEmailAddress;
+
+    @CCD(
+        label = "Defendant’s legal representative's email address",
```
scottstewart-sl: Minor - you've got different types of single quotes here.

charts/pcs-api/values.yaml
```diff
@@ -50,6 +50,7 @@ java:
     CCD_DATA_STORE_URL: "http://ccd-data-store-api-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal"
     CASE_DOCUMENT_AM_URL: "http://ccd-case-document-am-api-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal"
     RD_PROFESSIONAL_API_URL: "http://rd-professional-api-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal"
+    PCS_FRONTEND_URL: "https://pcs.aat.platform.hmcts.net"
```
paddy-hmcts: this will fail in other envs like prod, etc, you should probably use the `.Values` as above

charts/pcs-api/values.preview.template.yaml
```diff
@@ -39,6 +39,7 @@ java:
     CCD_SDK_DECENTRALISED: "true"
     PAYMENTS_API_URL: http://payment-api-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal
     PAY_CALLBACK_URL: https://${SERVICE_NAME}.preview.platform.hmcts.net/payment-update
+    PCS_FRONTEND_URL: ${PCS_FRONTEND_URL}
```
paddy-hmcts, line 54: should this not also be derived? It practically does nothing (as you're saying the env var passed in is equal to the env var)

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/LegalRepresentativeDetails.java
```diff
@@ -0,0 +1,64 @@
...
+        typeOverride = FieldType.Email
+    )
+    private String emailAddress;
+
+    @CCD(typeOverride = FieldType.Email)
+    private String originalEmailAddress;
+
+    @CCD(
+        label = "Defendant’s legal representative's reference",
```
paddy-hmcts: same as what @scottstewart-sl said above

src/main/java/uk/gov/hmcts/reform/pcs/ccd/PCSCaseView.java
```diff
@@ -154,6 +156,7 @@ private PCSCase getSubmittedCase(long caseReference) {
         caseLinkView.setCaseFields(pcsCase, pcsCaseEntity);
         caseNoteView.setCaseFields(pcsCase, pcsCaseEntity);
         flagsView.setCaseFields(pcsCase, pcsCaseEntity);
+        legalRepresentativeSummaryService.handleLegalRepresentativeSummary(pcsCase, pcsCaseEntity);
```
linusnorton: This method does a http call... we can't do that on every load surely?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/legalrepresentative/LegalRepresentativePageService.java
```diff
@@ -0,0 +1,125 @@
...
+            && legalRepresentativeDetails.getUseEmailAddress().equals(VerticalYesNo.NO)) {
+            legalRepresentativeOrganisationEntity.setEmail(legalRepresentativeDetails.getEmailAddress());
+        }
+
+        legalRepresentativeOrganisationEntity.setHasAmendedContactDetails(YesOrNo.YES);
+
+        legalRepresentativeOrganisationRepository.save(legalRepresentativeOrganisationEntity);
```
linusnorton: does this model need organisationId?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftCaseDataService.java
```diff
@@ -365,35 +353,37 @@ private DraftCaseDataEntity createNewDraft(long caseReference, EventId eventId,
...
-        UUID userId,
-        UUID partyId,
+        DraftCaseData draftCaseData,
         Supplier<Optional<DraftCaseDataEntity>> draftSupplier
     ) {
 
-        if (partyId != null) {
+        if (draftCaseData.getPartyId() != null) {
             log.info(
                 "Getting unsubmitted draft data: caseReference={}, eventId={}, userId={}, partyId={}",
```
linusnorton: userId should be orgId

src/main/resources/db/migration/V008__add_legal_representative_org_table.sql
```diff
@@ -0,0 +1,25 @@
+create TABLE legal_representative_org (
+  id uuid primary key,
+  organisation_id varchar(80),
+  case_id UUID NOT NULL REFERENCES public.pcs_case (id),
```
linusnorton: this and org_id should have an index

charts/pcs-api/values.preview.template.yaml
m014389b: I was setting it in the JenkinsFile_CNP LOC-175. 

The value would either come from the "pcs-frontend-pr" label or default to "https://pcs.aat.platform.hmcts.net/"

I've updated to cater for this now

charts/pcs-api/values.yaml
m014389b: done

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/LegalRepresentativeDetails.java
m014389b: done

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/LegalRepresentativeDetails.java
m014389b: done

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftCaseDataService.java
m014389b: done, and other instances here

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/legalrepresentative/LegalRepresentativePageService.java
m014389b: In theory the LegalRepresentativeOrganisationEntity passed in should always be present, changed to throw an exception instead

src/main/java/uk/gov/hmcts/reform/pcs/ccd/PCSCaseView.java
m014389b: We have https://tools.hmcts.net/jira/browse/HDPI-6650 in our board to fix this issue.

Theres a message on the engineering board discussing it also

charts/pcs-api/values.preview.template.yaml
m014389b: Side note: tested without it in this template it it defaults to the root template. So addd back in

src/main/java/uk/gov/hmcts/reform/pcs/ccd/PCSCaseView.java
tvr-hmcts: The problem here can be addressed differently ...  There is value in the HDPI-6650 but ...

Before this line ... in PCSCaseView.java — getSubmittedCase() makes 2 remote calls directly after each other for the same data before this line...

    L149: documentsView.setCaseFields(pcsCase, pcsCaseEntity)
        Call chain: [uk.gov](http://uk.gov/).hmcts.reform.pcs.ccd.view.GenAppsView.setCaseFields -> [uk.gov](http://uk.gov/).hmcts.reform.pcs.reference.service.OrganisationService.getOrganisationIdForCurrentUser -> [uk.gov](http://uk.gov/).hmcts.reform.pcs.reference.service.OrganisationDetailsService.getOrganisationIdentifier -> [uk.gov](http://uk.gov/).hmcts.reform.pcs.reference.service.OrganisationDetailsService.getOrganisationDetails -> [uk.gov](http://uk.gov/).hmcts.reform.pcs.reference.api.RdProfessionalApi.getOrganisationDetails


L159: genAppsView.setCaseFields(pcsCase, pcsCaseEntity)
        Call chain: [uk.gov](http://uk.gov/).hmcts.reform.pcs.ccd.view.GenAppsView.setCaseFields -> [uk.gov](http://uk.gov/).hmcts.reform.pcs.reference.service.OrganisationService.getOrganisationIdForCurrentUser -> [uk.gov](http://uk.gov/).hmcts.reform.pcs.reference.service.OrganisationDetailsService.getOrganisationIdentifier -> [uk.gov](http://uk.gov/).hmcts.reform.pcs.reference.service.OrganisationDetailsService.getOrganisationDetails -> [uk.gov](http://uk.gov/).hmcts.reform.pcs.reference.api.RdProfessionalApi.getOrganisationDetails


Both of these are above this 

legalRepresentativeSummaryService.handleLegalRepresentativeSummary(pcsCase, pcsCaseEntity);

All three call the same remote call.

So if you make this call:

String orgId = organisationService.getOrganisationIdForCurrentUser();

further up in this PCSCaseView.getSubmittedCase(long caseReference) method and then you can pass it into the rest.


The issue is visibility, code size and speed/number of PRs / context switches.  As these are not visible from within this PR window scope.

m014389b: Yeah that's a good point. I've moved the call up so will reduce the number of calls here by a factor at least

src/main/java/uk/gov/hmcts/reform/pcs/service/LegalRepresentativePartyLinkService.java
```diff
@@ -56,37 +54,28 @@ public void linkLegalRepresentativeToParty(long caseReference, String partyId, U
...
-            if (legalRepresentativeLinkedToCase) {
-                backfillOrganisationMetadata(legalRepresentative, user.getUid());
-                legalRepresentative.addParty(defendantPartyEntity);
-            } else {
-                legalRepresentative = createNewLegalRepresentative(organisationId,
-                                                                   idamId,
-                                                                   organisationService
-                                                                       .getNameAndAddress(user.getUid()));
-            }
+            backfillOrganisationMetadata(legalRepresentativeOrganisation, user.getUid());
```
tvr-hmcts: Line 44 of this method - linkLegalRepresentativeToParty calls to get the organisationId which this backfillOrganisationMetadata(...,...) also then calls internally.   Then also in turn there is another call for the NameAndAdress.

I think perhaps the NameAndAddress can be called at the start of the linkLegalRepresentativeToParty and then passed into these methods no?  I understand that it is cached so to speak but it is still a pull on the db pool each time when the values can probably just be pulled once instead.

Thinking that through a bit more it maybe worth extending that NameAndAddress record so that it carries the organisationId also

src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/legalrepresentative/OrganisationEntity.java
```diff
@@ -0,0 +1,104 @@
...
+    @JsonManagedReference
+    private List<PartyLegalRepresentativeOrganisationEntity> partyLegalRepresentativeOrganisationList =
+        new ArrayList<>();
+
+    @OneToMany(fetch = LAZY, cascade = ALL, mappedBy = "legalRepresentativeOrganisation")
+    @Builder.Default
+    @JsonManagedReference
+    private List<LegalRepresentativeEntity> legalRepresentativeList = new ArrayList<>();
+
+    public void addParty(PartyEntity party) {
```
tvr-hmcts, line 59: There needs to be coverage of this.  It is interesting as this nicely reduces the anemic nature of the domain / entity area.  Just needs a few tests :)

src/main/java/uk/gov/hmcts/reform/pcs/service/LegalRepresentativePartyLinkService.java
```diff
@@ -102,54 +91,62 @@ private void revokeDefendantAccessForRepresentedParty(long caseReference, PartyE
...
-            legalRepresentative.setOrganisationName(organisationService.getNameAndAddress(userId).name());
+        if (legalRepresentativeOrganisation.getOrganisationName() == null) {
+            legalRepresentativeOrganisation.setOrganisationName(organisationService
+                                                                    .getNameAndAddress(userId).name());
+        }
+    }
+
+    private void backfillLegalRepresentative(LegalRepresentativeOrganisationEntity legalRepresentativeOrganisation,
+                                             UUID idamId) {
+        boolean isLegalRepresentativeNotPresent = legalRepresentativeOrganisation.getLegalRepresentativeList().stream()
```
tvr-hmcts: Need a test for when this returns gets set to false.

src/main/java/uk/gov/hmcts/reform/pcs/service/LegalRepresentativePartyLinkService.java
```diff
@@ -167,24 +164,27 @@ private PartyEntity getDefendantPartyEntity(PcsCaseEntity caseEntity, String par
...
+        if (partyLinkedToLegalRepresentativeOrganisationAndActive.isPresent()) {
+            LegalRepresentativeOrganisationEntity existingLegalRepresentativeOrganisation =
+                partyLinkedToLegalRepresentativeOrganisationAndActive.get();
 
-            existingLegalRepresentative.getClaimPartyLegalRepresentativeList().stream()
-                .filter(claimPartyLegalRepresentative ->
-                            claimPartyLegalRepresentative.getParty().getId().equals(partyId))
-                .forEach(this::invalidateLegalRepresentativeClaimParty);
+            existingLegalRepresentativeOrganisation.getPartyLegalRepresentativeOrganisationList().stream()
+                .filter(partyLegalRepresentativeOrganisation ->
```
tvr-hmcts, line 198: Think there is a missing test when this filter does not return anything and the forEach then does not get called.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/legalrepcontactdetails/LegalRepresentativeContactDetails.java
```diff
@@ -0,0 +1,83 @@
...
+
+        new PageBuilder(eventBuilder)
+            .add(legalRepresentativeContactDetailsPage);
+    }
+
+    private PCSCase start(EventPayload<PCSCase, State> eventPayload) {
+        PCSCase pcsCase = eventPayload.caseData();
+        Long caseReference = eventPayload.caseReference();
+
+        String organisationId = organisationService.getOrganisationIdForCurrentUser();
```
tvr-hmcts, line 50: Just a thought as this is called in the submit also ... can this be set in the domain so that it does not need calling again on the submit?

tvr-hmcts (commented):
Part way through thoughts.

src/main/java/uk/gov/hmcts/reform/pcs/service/LegalRepresentativePartyLinkService.java
m014389b: added coverage

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/legalrepcontactdetails/LegalRepresentativeContactDetails.java
m014389b: added to the domain DTO

src/main/java/uk/gov/hmcts/reform/pcs/service/LegalRepresentativePartyLinkService.java
m014389b: added coverage

src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/legalrepresentative/OrganisationEntity.java
m014389b: added coverage

src/main/java/uk/gov/hmcts/reform/pcs/service/LegalRepresentativePartyLinkService.java
m014389b: Changed to cover the whole org details response we want. Renamed to OrganisationDetails to better reflect the intention

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/legalrepcontactdetails/LegalRepresentativeContactDetails.java
tvr-hmcts: Two thoughts 

1) Is the id all that is needed to be stored in the domain?   Is there any need to then recall for the rest in the flows?

2) Is that approach possible within the PCSCase also for the ResumePossessionClaim - where there are multiple calls - one in the start and one in the submit - it is the submit one that troubles me?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/ResumePossessionClaim.java
```diff
@@ -98,11 +98,11 @@ PCSCase start(EventPayload<PCSCase, State> eventPayload) {
 
         setUnsubmittedCaseDataFlag(caseReference, caseData);
 
-        NameAndAddress nameAndAddress = organisationService.getNameAndAddressForCurrentUser();
+        OrganisationDetails organisationDetails = organisationService.getOrganisationDetailsForCurrentUser();
```
tvr-hmcts: Possible to hold in the domain also here to remove the recall in the submit?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/genapp/StartEventHandler.java
```diff
@@ -38,8 +36,8 @@ public PCSCase start(EventPayload<PCSCase, State> eventPayload) {
 
     // Set represented parties if the current user is a legal rep
     private void setRepresentedParties(long caseReference, PCSCase caseData) {
-        UUID currentUserId = securityContextService.getCurrentUserId();
-        legalRepresentativeService.getRepresentedPartiesDynamicList(currentUserId, caseReference)
+        String organisationIdForCurrentUser = organisationService.getOrganisationIdForCurrentUser();
```
tvr-hmcts, line 40: Can this one be held in the domain?   It is called again in the getApplicantParty() of the SubmitEventHandler which also has another three remote calls in it (unrelated to organisation though)

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/genapp/GenAppVisibilityService.java
```diff
@@ -34,19 +33,19 @@ public boolean isGenAppVisibleToUser(GenAppEntity genAppEntity, UUID currentUser
         }
 
         PartyEntity applicantParty = genAppEntity.getParty();
-        if (applicantParty == null || currentUserId == null) {
+        if (applicantParty == null || organisationId == null) {
             return false;
```
tvr-hmcts, line 68: The tests do not hit this.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/genapp/GenAppVisibilityService.java
```diff
@@ -34,19 +33,19 @@ public boolean isGenAppVisibleToUser(GenAppEntity genAppEntity, UUID currentUser
...
-        return legalRepresentativeRepository
-            .isLegalRepresentativeLinkedToPartyAndActive(currentUserId, applicantParty.getId());
+        return legalRepresentativeOrganisationRepository
+            .isRepresentativeOrganisationLinkedToPartyAndActive(organisationId, applicantParty.getId());
     }
 
-    public List<GenAppEntity> getVisibleGenAppsToUser(Collection<GenAppEntity> genApps, UUID userId) {
+    public List<GenAppEntity> getVisibleGenAppsToUser(Collection<GenAppEntity> genApps, String organisationId) {
         if (genApps == null || genApps.isEmpty()) {
             return List.of();
```
tvr-hmcts, line 101: test missed

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/dashboard/task/ApplicationsTaskGroupEvaluator.java
```diff
@@ -18,14 +18,14 @@
 @Component
 public class ApplicationsTaskGroupEvaluator implements TaskGroupEvaluator {
 
-    private final SecurityContextService securityContextService;
+    private final OrganisationService organisationService;
     private final GenAppVisibilityService genAppVisibilityService;
 
     public ApplicationsTaskGroupEvaluator(
```
tvr-hmcts, line 27: Potential @AllArgsConstructor instead?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/genapp/GenAppVisibilityService.java
```diff
@@ -57,7 +56,7 @@ public List<GenAppEntity> getVisibleGenAppsToUser(Collection<GenAppEntity> genAp
                 GenAppEntity::getApplicationSubmittedDate,
                 Comparator.nullsLast(Comparator.reverseOrder())
             ))
-            .filter(genAppEntity -> isGenAppVisibleToUser(genAppEntity, userId))
+            .filter(genAppEntity -> isGenAppVisibleToUser(genAppEntity, organisationId))
```
tvr-hmcts: Wondering if there can be an optimisation here.  It is currently N hits in on the database.

src/main/java/uk/gov/hmcts/reform/pcs/service/LegalRepresentativePartyLinkService.java
tvr-hmcts: I see the test when this is false [linkLegalRepresentativeToParty_WithExistingLegalRepresentativeForOrg_SavesNewEntity()] but does it need another assert to make sure that the legalRepresentative is not added to the list?  Unless Im missing it.

Perhaps the method could be opened up (made package private) in order to make it easier to test (as the tests are quite large to setup).  Ive see use of @VisibleForTesting used in other projects.  Better though maybe extracting out certain logic - though perhaps that is for another day.

tvr-hmcts (commented):
Thoughts.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/genapp/GenAppVisibilityService.java
m014389b: I would defer this as I haven't touched this area and this PR has already had a fair amount of scope creep.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/genapp/GenAppVisibilityService.java
m014389b: I would defer this as I haven't touched this area and this PR has already had a fair amount of scope creep.

tvr-hmcts: Yep ok.   Can be revisited if the loop around on the db is a performance issue.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/genapp/GenAppVisibilityService.java
tvr-hmcts: Yep sorry over zealous

tvr-hmcts (approved):

tvr-hmcts (commented):
One note - Sorry should have put this on when approving.

- Feature flags.  - Not sure how that fits with this one.

src/main/resources/db/migration/V008__add_legal_representative_org_table.sql
```diff
@@ -0,0 +1,27 @@
...
+  organisation_name varchar(120),
+  email varchar(120),
+  phone varchar(40),
+  contact_reference varchar(80),
+  address_id  integer REFERENCES address (id),
+  has_amended_contact_details YES_NO
+);
+
+create table party_legal_rep_org (
```
m014389b: rename to claim_party_legal_representative_org

src/main/resources/db/migration/V008__add_legal_representative_org_table.sql
```diff
@@ -0,0 +1,27 @@
+create TABLE legal_representative_org (
+  id uuid primary key,
+  organisation_id varchar(80),
+  case_id UUID NOT NULL REFERENCES public.pcs_case (id),
+  organisation_name varchar(120),
+  email varchar(120),
+  phone varchar(40),
+  contact_reference varchar(80),
+  address_id  integer REFERENCES address (id),
+  has_amended_contact_details YES_NO
```
m014389b: Need to drop and rework how we know if LR details have been amended

src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/legalrepresentative/LegalRepresentativeOrganisationEntity.java
```diff
@@ -0,0 +1,79 @@
...
+    private Integer id;
+
+    private String organisationName;
+
+    private String organisationId;
+
+    private String organisationProfileId;
+
+    @OneToOne(mappedBy = "legalRepresentativeOrganisation", cascade = ALL)
+    private LegalRepresentativeOrganisationContactDetailsEntity legalRepresentativeOrganisationContactDetails;
```
jonathanditcher-solirius: is this meant to be a list, i.e. one lro to many case+lro contact detials

m014389b: No, previously when email phone etc was on the LRO it was just the latest contact details. No need to keep a history of contact detail changes

jonathanditcher-solirius: oh I see is that because the LRO is being retrieved via a caseId hence will retrieve the contact details applicable to that LRO + caseId

PR #1850, opened 2026-05-19
HDPI-5493 - Create Document Type list dynamically for England and Wales specific

guygrewal77:
### Jira link

See [HDPI-5493](https://tools.hmcts.net/jira/browse/HDPI-5493)

### Change description

The Document List will be different for England and Wales and so will be created dynamically in the Upload Additional documents page.

To ensure any previously entered values are retained, a call is made to DraftDataService to populate the AdditionalDocuments. 

### Testing done
XUI testing to ensure the Document Types are in the right order and country specific, and also the documents are placed in the correct folders in Case File View.

Screenshot 2026-05-20 at 09 29 05
Screenshot 2026-05-20 at 09 35 05
Screenshot 2026-05-20 at 10 48 59
Screenshot 2026-05-20 at 09 47 17

  * [ ] Yes
  * [x] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
```diff
@@ -59,7 +59,7 @@ public List<DocumentEntity> createAllDocuments(EnforcementOrder enforcementOrder
     private List<DocumentHolder> getPcsCaseDocuments(PCSCase pcsCase) {
         List<DocumentHolder> allDocuments = new ArrayList<>();
 
-        allDocuments.addAll(mapAdditionalDocumentsWithType(pcsCase.getAdditionalDocuments()));
+        allDocuments.addAll(mapAdditionalDocumentsWithType(pcsCase.getAdditionalDocs()));
```
sadmanrahh: For this new additional-docs path, AC09 says uploaded documents should be renamed to [user generated name] - [Claimant 1].pdf. I think we still only persist the original filename as the display name later in createDocumentEntities. Could we apply the claimant suffix and add a test for it?

sadmanrahh: Should the fixed document-type mappings match the ACs here? AC02 says Notice details uploads should be `POSSESSION_NOTICE` so they appear under Property documents, but they’re currently mapped as `NOTICE_FOR_SERVICE_OUT_OF_JURISDICTION`. AC03 says Tenancy/licence details uploads should be `TENANCY_AGREEMENT`, but they’re currently mapped as `TENANCY_LICENCE`.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/resumepossessionclaim/WantToUploadDocuments.java
```diff
@@ -1,19 +1,72 @@
...
     }
+
+    private AboutToStartOrSubmitResponse<PCSCase, State> midEvent(CaseDetails<PCSCase, State> details,
+                                                                  CaseDetails<PCSCase, State> detailsBefore) {
+        PCSCase caseData = details.getData();
+        AdditionalDocuments additionalDocuments = new AdditionalDocuments();
+        LegislativeCountry legislativeCountry = caseData.getLegislativeCountry();
+
+        additionalDocuments.setDocumentTypeList(createAdditionalDocumentList(legislativeCountry));
+        caseData.setAdditionalDocs(new ArrayList<>());
```
sadmanrahh: Could we add a test for selecting NO here? The mid-event currently initialises additionalDocs regardless of the answer, so a test would confirm no blank upload row is created when the user does not want to upload documents.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
guygrewal77: Fixed.

guygrewal77: Implemented now.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/resumepossessionclaim/WantToUploadDocuments.java
guygrewal77: Fixed.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
sadmanrahh: Looks good, thank you

sadmanrahh: Great, looks good. One quick question, In createDefendantUploadedDocuments() (line 220), defendant documents use the original filename without the suffix. Is that intentional, or should those also get the claimant suffix applied?

guygrewal77: That's part of the respond to a claim so we can ignore that here.

sadmanrahh: Great, thanks

sadmanrahh (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/AdditionalDocumentType.java
```diff
@@ -24,4 +35,19 @@ public enum AdditionalDocumentType implements HasLabel {
...
     private final String label;
 
+    public static AdditionalDocumentType getValueFromLabel(String label) {
+        return Arrays.stream(values()).filter(v -> v.getLabel().equals(label)).findFirst()
+            .orElseThrow(() -> new IllegalArgumentException("No AdditionalDocumentType with label: " + label));
+    }
+
+    public static boolean isEnglandSpecific(AdditionalDocumentType val) {
+        return val == TENANCY_AGREEMENT;
+    }
```
scottstewart-sl: I think a more elegant way to achieve this is for the enum constructor to take a Set of the legislative countries for which it applies. See `ClaimantType` for an example.

It also makes it easier to build up the dynamic list to show based on the legislative country.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/PCSCase.java
```diff
@@ -427,12 +427,19 @@ public class PCSCase {
...
         label = "Add document",
         hint = "Upload a document to the system"
     )
     private List<ListValue<AdditionalDocument>> additionalDocuments;
 
+    @CCD(
+        label = "Add document",
+        hint = "Upload a document to the system"
+    )
+    private List<ListValue<AdditionalDocuments>> additionalDocs;
```
scottstewart-sl: This doesn't feel very clean to have another field which has a similar name to the one above. I presume you're only doing it to prevent an issue with the ES index but I think at this stage it would be better to make the change properly and be ready to drop the ES index since we're not in Prod yet.

Also it would be better to re-use the existing `AdditionalDocument` class, rather than making an `AdditionalDocuments` class with an 's', even though it still only contains one document 🙂

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
```diff
@@ -162,10 +165,25 @@ private List<DocumentEntity> createDocumentEntities(
...
     }
 
+    private String getFilename(String uploadedFilename) {
+        String fileExtension = "";
+        String fileName = uploadedFilename;
+        int lastDot = uploadedFilename.lastIndexOf('.');
+        if (lastDot >= 0 && lastDot < uploadedFilename.length() - 1) {
+            fileExtension = uploadedFilename.substring(lastDot + 1);
+            fileName = uploadedFilename.substring(0, lastDot);
+        }
```
scottstewart-sl: You can use `FilenameUtils.getBaseName` and `FilenameUtils.getExtension` here to save having to implement it yourself.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/PCSCase.java
```diff
@@ -547,6 +554,7 @@ public class PCSCase {
     private ComponentLauncher caseFileView;
 
     @CCD(searchable = false)
+
```
tvr-hmcts: Minor vertical spacing?

tvr-hmcts (approved):
LGTM now.  Just a minor and I saw Scotts comment while looking at this which I think is worth the consideration.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
guygrewal77: Changed now to use the library.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/PCSCase.java
guygrewal77: Removed AdditionalDocuments class now.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
```diff
@@ -162,10 +166,19 @@ private List<DocumentEntity> createDocumentEntities(
                 .toList();
     }
 
+    private String getFilename(String uploadedFilename) {
+        return FilenameUtils.getBaseName(uploadedFilename) + " - " + CLAIMANT_1
+                + FilenameUtils.getExtension(uploadedFilename);
+    }
```
scottstewart-sl, line 207: It's not ideal to assume that any time this DocumentService.createAllDocuments(pcsCase) method is called, it is for a claimant, (even if that is the only way it is used at the moment).

It would have been better to derive the party label in a similar way to `DocumentNameService` on one of my branches, like I mentioned to you on the chat the other day. However, I realise that would require a bit of other refactoring to be able to determine which party the current user represents, (by IDAM ID), and what their rank is.

So it's OK to leave for now, but I suspect we may need to revisit this in the future.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
```diff
@@ -118,7 +121,8 @@ private List<DocumentHolder> mapAdditionalDocumentsWithType(
         return ListValueUtils.unwrapListItems(documents).stream()
             .map(doc -> DocumentHolder.builder()
                 .document(doc.getDocument())
-                .type(mapAdditionalDocumentTypeToDocumentType(doc.getDocumentType()))
+                .type(mapAdditionalDocumentTypeToDocumentType(
+                        AdditionalDocumentType.getValueFromLabel(doc.getDocumentType().getValueLabel())))
```
scottstewart-sl, line 160: I might be missing something, but isn't `doc.getDocumentType()` already an `AdditionalDocumentType` so this code is just converting it to a label and back again?

guygrewal77: Dynamic List has a UUID as a key so we need to get the enum's key from the label to retrieve the enum value. We don't have the enum's key stored as a String anymore.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
guygrewal77: There's scope for quite a lot of refactoring in other areas too, e.g. respond to a claim functionality is not easily distinguishable.

I agree this can be left for the next release.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/PCSCase.java
guygrewal77: Fixed.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/AdditionalDocumentType.java
guygrewal77: Fixed.

tvr-hmcts (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
```diff
@@ -163,10 +167,19 @@ private List<DocumentEntity> createDocumentEntities(
                 .toList();
     }
 
+    private String getFilename(String uploadedFilename) {
+        return FilenameUtils.getBaseName(uploadedFilename) + " - " + CLAIMANT_1
+                + FilenameUtils.getExtension(uploadedFilename);
```
sadmanrahh: This drops the dot before the file extension. FilenameUtils.getExtension("file.pdf") returns pdf, so this becomes file - Claimant 1pdf. Could we add "." + extension when an extension is present, and add a test for filename.pdf?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
```diff
@@ -152,7 +156,7 @@ private List<DocumentEntity> createDocumentEntities(
                 .map(holder -> DocumentEntity.builder()
                         .url(holder.getDocument().getUrl())
                         .documentId(documentIdExtractor.extractDocumentId(holder.getDocument().getUrl()))
-                        .fileName(holder.getDocument().getFilename())
+                        .fileName(getFilename(holder.getDocument().getFilename()))
```
sadmanrahh: This renames every document created through createDocumentEntities(), including enforcement documents. AC09 is only for Make a Claim uploads, so could we scope the renaming to the PCSCase path and keep enforcement filenames unchanged?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/AdditionalDocumentType.java
```diff
@@ -1,27 +1,56 @@
...
+    WITNESS_STATEMENT("Witness statement", Set.of(ENGLAND, WALES)),
+    RENT_STATEMENT("Rent statement", Set.of(ENGLAND, WALES)),
+
+    // England specific types
+    TENANCY_AGREEMENT("Tenancy agreement", Set.of(ENGLAND)),
+
+    // Wales specific types
+    OCCUPATION_LICENCE("Occupation contract or licence", Set.of(WALES)),
+    ENERGY_PERFORMANCE_CERTIFICATE("Energy performance certificate", Set.of(WALES)),
+    GAS_SAFETY_CERTIFICATE("Gas safety certificate", Set.of(WALES)),
```
sadmanrahh, line 26: Minor - AC07 refers to Gas safety report, but this label is Gas safety certificate.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/DocumentType.java
```diff
@@ -9,7 +9,10 @@
 public enum DocumentType implements HasLabel {
 
     RENT_STATEMENT("Rent statement"),
-    OCCUPATION_LICENCE("Occupation licence"),
+    OCCUPATION_LICENCE("Occupation contract or licence"),
+    ENERGY_PERFORMANCE_CERTIFICATE("Energy performance certificate"),
+    GAS_SAFETY_CERTIFICATE("Gas safety certificate"),
```
sadmanrahh, line 14: Minor- same thing, might have to be changes to Gas safety report, based on AC07

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
guygrewal77: Good point.

guygrewal77: Fixed.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
guygrewal77: Fixed

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/AdditionalDocumentType.java
guygrewal77: I think it should be certificate - in the table it is so it might be an error in the AC.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/DocumentType.java
guygrewal77: I think it should be certificate - in the table it is so it might be an error in the AC.

sadmanrahh (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/resumepossessionclaim/UploadAdditionalDocumentsDetails.java
```diff
@@ -64,13 +63,13 @@ private AboutToStartOrSubmitResponse<PCSCase, State> midEvent(CaseDetails<PCSCas
     }
 
     public List<String> validateDocumentDescription(
-        List<ListValue<AdditionalDocument>> additionalDocs,
+        List<ListValue<AdditionalDocument>> additionalDocuments,
```
tvr-hmcts, line 66: Any idea why the method is public?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/AdditionalDocumentType.java
```diff
@@ -1,27 +1,56 @@
...
+import java.util.Arrays;
+import java.util.Set;
+
+import static java.util.Objects.requireNonNull;
+import static uk.gov.hmcts.reform.pcs.postcodecourt.model.LegislativeCountry.ENGLAND;
+import static uk.gov.hmcts.reform.pcs.postcodecourt.model.LegislativeCountry.WALES;
 
-@AllArgsConstructor
 @Getter
 public enum AdditionalDocumentType implements HasLabel {
```
tvr-hmcts, line 15: Think we need a test of its' own right for all the public methods.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/resumepossessionclaim/WantToUploadDocuments.java
```diff
@@ -1,19 +1,91 @@
...
+        PCSCase caseData = details.getData();
+
+        setAdditionalDocumentsFromDraft(details.getId(), caseData);
+
+        if (caseData.getWantToUploadDocuments().equals(VerticalYesNo.YES)
+            && CollectionUtils.isEmpty(caseData.getAdditionalDocuments())) {
+            AdditionalDocument additionalDocuments = new AdditionalDocument();
+            LegislativeCountry legislativeCountry = caseData.getLegislativeCountry();
+            additionalDocuments.setDocumentType(createAdditionalDocumentList(legislativeCountry));
+            caseData.setAdditionalDocuments(new ArrayList<>());
```
gmmagruder, line 54: Should caseData.setAdditionalDocuments(new ArrayList<>()) be set outside of this if statement in case the user changes the answer from Yes to No?

src/test/java/uk/gov/hmcts/reform/pcs/ccd/page/resumepossessionclaim/WantToUploadDocumentsTest.java
```diff
@@ -0,0 +1,229 @@
...
+import java.util.UUID;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+class WantToUploadDocumentsTest extends BasePageTest {
```
tvr-hmcts, line 36: Nice :)

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/resumepossessionclaim/WantToUploadDocuments.java
```diff
@@ -1,19 +1,91 @@
...
             .label("wantToUploadDocuments-separator", "---")
             .mandatory(PCSCase::getWantToUploadDocuments)
             .label("wantToUploadDocuments-saveAndReturn", CommonPageContent.SAVE_AND_RETURN);
     }
+
+    private AboutToStartOrSubmitResponse<PCSCase, State> midEvent(CaseDetails<PCSCase, State> details,
+                                                                  CaseDetails<PCSCase, State> detailsBefore) {
+        PCSCase caseData = details.getData();
+
+        setAdditionalDocumentsFromDraft(details.getId(), caseData);
```
gmmagruder, line 47: What is the purpose of this? If the user answers yes then the additional documents gets reset. If the user answers no, it seems like there shouldn't be any documents set.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
```diff
@@ -72,7 +85,7 @@ private List<DocumentHolder> getPcsCaseDocuments(PCSCase pcsCase) {
         allDocuments.addAll(mapDocumentsWithType(
             Optional.ofNullable(pcsCase.getTenancyLicenceDetails())
                     .map(TenancyLicenceDetails::getTenancyLicenceDocuments)
-                    .orElse(null), DocumentType.TENANCY_LICENCE));
+                    .orElse(null), DocumentType.TENANCY_AGREEMENT));
```
gmmagruder, line 95: Is the document still visible in the case details tab? Also is it possible for there to be other documents with the TENANCY_AGREEMENT category? This might cause the unintended affect of displaying unintended documents.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
```diff
@@ -82,7 +95,7 @@ private List<DocumentHolder> getPcsCaseDocuments(PCSCase pcsCase) {
         allDocuments.addAll(mapDocumentsWithType(
             Optional.ofNullable(pcsCase.getNoticeServedDetails())
                     .map(NoticeServedDetails::getNoticeDocuments)
-                    .orElse(null), DocumentType.NOTICE_FOR_SERVICE_OUT_OF_JURISDICTION));
+                    .orElse(null), DocumentType.POSSESSION_NOTICE));
```
gmmagruder, line 105: Same as above but for POSSESSION_NOTICE

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentServiceTest.java
```diff
@@ -947,17 +1022,17 @@ void shouldFilterOutNullValuesFromCounterClaimDocuments() {
...
     }
 
     private static WalesDocuments buildRequiredDocumentsWales(DocumentType documentType,
                                                               List<ListValue<Document>> documents) {
         return switch (documentType) {
             case ENERGY_PERFORMANCE_CERTIFICATE -> WalesDocuments.builder().energyPerformance(documents).build();
-            case GAS_SAFETY_REPORT -> WalesDocuments.builder().gasSafetyReport(documents).build();
-            case ELECTRICAL_INSTALLATION_CONDITION ->
+            case GAS_SAFETY_CERTIFICATE -> WalesDocuments.builder().gasSafetyReport(documents).build();
+            case EICR_REPORT ->
```
tvr-hmcts, line 1307: Nothing hits the default type behaviour at the moment.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/resumepossessionclaim/WantToUploadDocuments.java
tvr-hmcts: Would there not be a scenario where they originally set something and have now updated it?

Though would the case have not been inflated already correctly already?

tvr-hmcts (commented):
Some comments.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/resumepossessionclaim/WantToUploadDocuments.java
guygrewal77: This is needed to get the saved answers if the user decides to go back to the previous pages after getting past the UploadAdditionalDocuments page. For some reason, the document list was getting lost and the only way to re-populate it is to fetch the data back from draft.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
```diff
@@ -180,10 +190,23 @@ private List<DocumentEntity> createDocumentEntities(
                 .toList();
     }
 
+    private void applyClaimFilename(List<DocumentHolder> allDocuments) {
+        allDocuments.forEach(dh -> {
+            String uploadedFilename = dh.getDocument().getFilename();
+            dh.getDocument().setFilename(FilenameUtils.getBaseName(uploadedFilename) + " - " + CLAIMANT_1
```
sadmanrahh, line 203: AC09 says the renamed file should use the user-generated name, but this uses the original uploaded filename base. For additional docs, should this use the entered description/name when present, e.g. Photos of damp in bedroom - Claimant 1.pdf, and only fall back to the original filename if no user name exists?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/resumepossessionclaim/WantToUploadDocuments.java
guygrewal77: If the user changes to No, then we don't want to proceed to the next page - UploadAdditionalDocuments so this object won't get created.

It is only created for the first time when the user first visits this page at which point the list will be empty.

If the user proceeds to CYA after uploading documents and then decides to revisit the WantToUploadDocuments page, we don't want to create a new AdditionalDocument.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
guygrewal77: The Tenancy agreement document can be uploaded in the Tenancy details page and also in the Upload Additional documents page. The DocumentService doesn't differentiate between where the document was uploaded as it's simply persisting all the documents uploaded throughout the journey.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
guygrewal77: As above, this is persisting all documents uploaded throughout the MAC journey.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
gmmagruder: So this might cause a problem with the case tabs since it fetches the document to display based on the document type. If documents outside of the tenancy page can have the same document type, this will cause both documents to be displayed in the case details tab where only the one uploaded on the tenancy page is meant to be displayed.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/AdditionalDocumentType.java
guygrewal77: There is coverage for all the public methods in the enum. Might be overkill to have designated tests for enums though ?

tvr-hmcts: Is an enum with business logic though.   We dance on the edge of the coverage levels in the pipeline.

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentServiceTest.java
guygrewal77: I don't see the point in testing the default case here ? Maybe it should just return a null instead of throwing an exception ? 

The tests are already testing for a valid document type.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/AdditionalDocumentType.java
guygrewal77: It's got coverage already

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/resumepossessionclaim/UploadAdditionalDocumentsDetails.java
guygrewal77: fixed

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
guygrewal77: No, it should be what's in the filename. I did clarify on this as I was under the same impression before of using the description field.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentService.java
guygrewal77: The only way to differentiate between the two is by using the description field. Only the additional documents have that field populated.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/NoticeOfPossessionView.java
```diff
@@ -95,7 +95,7 @@ private static List<ListValue<Document>> getNoticeStatement(PcsCaseEntity pcsCas
     }
 
     private static boolean isNoticeStatement(DocumentEntity documentEntity) {
-        return documentEntity.getType() == DocumentType.NOTICE_FOR_SERVICE_OUT_OF_JURISDICTION;
+        return documentEntity.getType() == DocumentType.POSSESSION_NOTICE;
```
gmmagruder, line 104: Can you also make sure the filter by whether it has a description as discussed.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/TenancyLicenceView.java
```diff
@@ -78,7 +78,7 @@ private static List<ListValue<Document>> getTenancyLicenceDocument(PcsCaseEntity
     }
 
     private static boolean isTenancyLicence(DocumentEntity documentEntity) {
-        return documentEntity.getType() == DocumentType.TENANCY_LICENCE;
+        return documentEntity.getType() == DocumentType.TENANCY_AGREEMENT;
```
gmmagruder, line 83: Can you also make sure the filter by whether it has a description as discussed.

gmmagruder (commented):
Can you also filter rent statements in RentArrearsView on whether it has an description

src/test/java/uk/gov/hmcts/reform/pcs/ccd/view/NoticeOfPossessionViewTest.java
```diff
@@ -296,4 +298,57 @@ void shouldSetNoticeDocumentIfPresent() {
...
+        LocalDate postedDate = mock(LocalDate.class);
+        UUID noticeDocumentId = UUID.randomUUID();
+
+        when(noticeOfPossessionEntity.getServingMethod()).thenReturn(FIRST_CLASS_POST);
+        when(noticeOfPossessionEntity.getNoticeDate()).thenReturn(postedDate);
+        when(pcsCaseEntity.getDocuments()).thenReturn(
+                List.of(
+                        DocumentEntity.builder()
+                                .id(noticeDocumentId)
+                                .type(DocumentType.POSSESSION_NOTICE)
```
gmmagruder, line 369: Missing document of POSSESSION_NOTICE with description

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/DocumentsView.java
```diff
@@ -62,4 +63,7 @@ public boolean isDocumentVisibleToUser(DocumentEntity documentEntity, UUID curre
         }
     }
 
+    public static boolean isDescriptionEmpty(DocumentEntity documentEntity) {
```
gmmagruder, line 66: Missing unit test for new method (I know it is tested in other unit tests but we should ensure that this is tested in isolation)

guygrewal77: Fixed now

src/test/java/uk/gov/hmcts/reform/pcs/ccd/view/NoticeOfPossessionViewTest.java
guygrewal77: Fixed now.

gmmagruder (approved):

PR #1839, opened 2026-05-16
HDPI-6031: Gen apps supporting document upload screen

scottstewart-sl:
### Jira link

See [HDPI-6031](https://tools.hmcts.net/jira/browse/HDPI-6031)

### Change description

Add screens for uploading supporting documents for a Gen App.

### Testing done

Manual testing locally

### Security Vulnerability Assessment ###

**CVE Suppression:** Are there any CVEs present in the codebase (either newly introduced or pre-existing) that are being intentionally suppressed or ignored by this commit?
  * [ ] Yes
  * [x] No

### Checklist

- [x] commit messages are meaningful and follow good commit message guidelines
- [ ] README and other documentation has been updated / added (if needed)
- [x] tests have been updated / new tests has been added (if needed)
- [ ] Does this PR introduce a breaking change

PR #1853, opened 2026-05-19
HDPI-6510: Gen Apps supporting docs

scottstewart-sl:
### Jira link

See [HDPI-6510](https://tools.hmcts.net/jira/browse/HDPI-6510)

### Change description

Rename supporting documents that were uploaded as part of a citizen gen app, to include the gen app number and defendant label as a postfix before the file extension.

Related pcs-frontend PR: https://github.com/hmcts/pcs-frontend/pull/1381

### Testing done

Local testing with corresponding pcs-frontend changes

  * [ ] Yes
  * [x] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/genapp/GenAppService.java
```diff
@@ -51,6 +58,10 @@ public GenAppEntity createGenAppEntity(CitizenGenAppRequest citizenCreateGenApp,
             .appliedForHwf(citizenCreateGenApp.getAppliedForHwf())
             .build();
 
+        // Adding the Gen App to the PcsCaseEntity allocates it a rank,
+        // which we rely on later on in this method to rename the supporting documents
+        pcsCaseEntity.addGenApp(genAppEntity);
```
tvr-hmcts, line 63: Interesting.   Is there a reason why more business logic is starting to be part of the Entity now?   Typically these are anemic but now they are becoming enriched with domain knowledge so I'm just curious on the approach.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentNameService.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        if (!extension.isBlank()) {
+            filename += "." + extension;
+        }
+
+        return filename;
+    }
+
+    private static String getPartyLabel(ClaimEntity mainClaim, UUID partyId) {
```
tvr-hmcts, line 42: static?

Also if this can return null then would an Optional<String> return value be better?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentNameService.java
```diff
@@ -0,0 +1,61 @@
...
+        if (applicantClaimParty.getRole() == PartyRole.CLAIMANT) {
+            return "Claimant %d".formatted(applicantClaimParty.getRank());
+        } else if (applicantClaimParty.getRole() == PartyRole.DEFENDANT) {
+            return "Defendant %d".formatted(applicantClaimParty.getRank());
+        } else {
+            return null;
+        }
+    }
+
+    private static ClaimPartyEntity getClaimParty(ClaimEntity claim, UUID partyId) {
```
tvr-hmcts, line 54: Not sure why static?

tvr-hmcts (commented):
Just some quesions from me.

guygrewal77:
Not related to your changes but is it possible to get some coverage on GenAppDocumentGenerator.getPartyLabel() to cover all scenarios ?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentNameService.java
scottstewart-sl: My IDE added the static when I refactored the code out into a method, but I think it's OK to keep it isn't it, as the method doesn't rely on any class fields.

I could return Optional, but this is a private method only used by this class and I think it would make the code slightly more complex as I'd have to wrap the return values in `Optional.of()`

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentNameService.java
scottstewart-sl: Same as above - it doesn't reference any class fields

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/genapp/GenAppService.java
scottstewart-sl: Yes, I agree that we should avoid domain logic in the entities where possible. This is following the pattern of getting the parent entity to add a child to a list itself, so that it can set the appropriate field in the child at the same time for the 2 way association.

The requirement here is to know which number in the list this particular gen app entity is, and we can't know that until it is added. It's not really a lot of domain knowledge in this case, just knowing what position in the list the gen app is.

scottstewart-sl:
@guygrewal77 

> Not related to your changes but is it possible to get some coverage on GenAppDocumentGenerator.getPartyLabel() to cover all scenarios ?

I've actually refactored this code out in my next PR, (still in draft), and that adds coverage for that method, (after it was moved):

See `DocumentNamingServiceTest.java` in 

https://github.com/hmcts/pcs-api/pull/1839/changes#diff-f85ea874cddd37a3d372bd2c30b1ef38c3ff4b26fd1237b503e45feb409d6916

tvr-hmcts (approved):

aqibh25 (approved):

PR #1930, opened 2026-06-04
Add Work Allocation preview chart to pcs-api

linusnorton:
Adds an opt-in Work Allocation stack to pcs-api preview environments, selected by the **\`pr-values:wa\`** GitHub label (stacked on top of \`pr-values:ccd\`).


damongreen123 (approved):

bin/setup-role-assignments
```diff
@@ -1,28 +1,67 @@
 #!/usr/bin/env bash
 set -eu
```
scottstewart-sl, line 3: @linusnorton What's the purpose of this script? Wouldn't the users already have the correct roles in AAT IDAM, (or cftlib-am-role-assignments.json for local development)?

bin/wa/import-dmn-diagrams.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
```
scottstewart-sl, line 1: Minor - it would be slightly more accurate to name this file in plural, e.g.

`import-dmn-diagrams.sh`

Jenkinsfile_CNP
```diff
@@ -166,6 +175,32 @@ withPipeline(type, product, component) {
     env.CCD_ENABLED = "true"
   }
 
+  afterSuccess('akschartsinstall') {
+    onPR {
+      for (label in githubApi.getLabelsbyPattern(env.BRANCH_NAME, "pr-values:wa") ) {
```
scottstewart-sl: Wouldn't this be better as an `if` as we would only be expecting a maximum of one match?

bin/setup-role-assignments
linusnorton: I've taken this out of this PR but I believe the script is still there for local dev

scottstewart-sl (approved):

PR #1934, opened 2026-06-05
HDPI-6833: add fields to case list view

gmmagruder:
### Jira link

See [HDPI-6833](https://tools.hmcts.net/jira/browse/HDPI-6833)

### Change description

Add fields to case list

Screenshot 2026-06-05 at 10 10 16

Screenshot 2026-06-05 at 10 50 46

### Testing done

  * [ ] Yes
  * [x] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseListView.java
```diff
@@ -0,0 +1,80 @@
...
+        setDefendantNames(pcsCase);
+        setDateIssuedString(pcsCase);
+        setPostCode(pcsCase);
+    }
+
+    private void setClaimantNames(PCSCase pcsCase) {
+        List<ListValue<Party>> claimants = pcsCase.getAllClaimants();
+
+        if (!CollectionUtils.isEmpty(claimants)) {
+            pcsCase.setClaimantNames(claimants.getFirst().getValue().getOrgName());
```
toby-plunkett, line 33: This does imply that all claimants are from the same Organisation which _I _think__ is right, but I don't know if there is a case where it might not be? Maybe worth double checking

toby-plunkett (approved):
Just that one comment to check

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseListView.java
gmmagruder: We only have one claimant at the moment and the claimant's name is stored under orgName. In future if we have multiple claimants, we would have to implement logic similar to how the defendant names are made.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseListView.java
```diff
@@ -0,0 +1,80 @@
...
+import uk.gov.hmcts.reform.pcs.ccd.domain.VerticalYesNo;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.List;
+
+@Component
+public class CaseListView {
+
+    private static final String PERSON_UNKNOWN = "Persons Unknown";
```
sadmanrahh: AC04 mentions it in lower case 'persons unknown'. Can you please confirm this and align accordingly?

src/test/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseListViewTest.java
```diff
@@ -0,0 +1,167 @@
...
+            .allDefendants(List.of(defendant1ListValue, defendant2ListValue, defendant3ListValue))
+            .build();
+
+        // When
+        caseListView.setCaseFields(pcsCase);
+
+        // Then
+        assertThat(pcsCase.getDefendantNames()).isEqualTo("One, Two and Others");
+    }
+
```
sadmanrahh, line 101: Can we add coverage for the mixed known/unknown defendant scenarios from AC04, e.g. "persons unknown, Jackson" and "Jackson, persons unknown"?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/ClaimService.java
```diff
@@ -88,4 +91,9 @@ private ClaimEntity buildClaimEntity(PCSCase pcsCase) {
             .build();
     }
 
+    public void setClaimIssuedDate(ClaimEntity claimEntity) {
+        claimEntity.setClaimIssuedDate(LocalDateTime.now(utcClock));
```
sadmanrahh: This will overwrite claimIssuedDate if the payment event runs again. Can we only set it when the existing value is null, so the original issued date remains stable?

gmmagruder: Can this ever be run multiple times? This feels like a system event that should only ever be run once.

sadmanrahh: It might be possible because the payment event is valid for both PENDING_CASE_ISSUED and CASE_ISSUED, and a retried PAID callback would call it again.

gmmagruder: Ok in that case, have made it so that claimIssuedDate is only set once.

src/test/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseListViewTest.java
gmmagruder: Addressed

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseListView.java
gmmagruder: Addressed

sadmanrahh (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseListView.java
```diff
@@ -0,0 +1,80 @@
...
+        if (defendant.getNameKnown() == VerticalYesNo.YES) {
+            return defendant.getLastName();
+        }
+
+        return PERSON_UNKNOWN;
+    }
+
+    private void setDateIssuedString(PCSCase pcsCase) {
+        LocalDateTime dateIssued = pcsCase.getDateIssued();
+        if (dateIssued == null) {
```
guygrewal77: Could move the assigning of the dateIssued field inside the IF instead and negating the condition.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseListView.java
```diff
@@ -0,0 +1,80 @@
...
+            return;
+        }
+
+        pcsCase.setDateIssuedString(dateIssued.format(DATE_FORMATTER));
+    }
+
+    private void setPostCode(PCSCase pcsCase) {
+        AddressUK address = pcsCase.getPropertyAddress();
+        if (address == null) {
```
guygrewal77: Shouldn't the postcode always be available as there wouldn't be a claim without a property address ?

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/event/PaymentEvent.java
```diff
@@ -32,6 +35,8 @@ public void configureDecentralised(DecentralisedConfigBuilder<PCSCase, State, Us
 
     private SubmitResponse<State> submit(EventPayload<PCSCase, State> eventPayload) {
         log.info("Received: {}", eventPayload);
+        long caseReference = eventPayload.caseReference();
+        pcsCaseService.setCaseIssuedDate(caseReference);
```
guygrewal77: Can extract the dateIssued from the PCSCase and only make the call if it's not set.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/ClaimService.java
```diff
@@ -88,4 +91,11 @@ private ClaimEntity buildClaimEntity(PCSCase pcsCase) {
             .build();
     }
 
+    public void setClaimIssuedDate(ClaimEntity claimEntity) {
+        if (claimEntity.getClaimIssuedDate() == null) {
```
guygrewal77: This check is not needed as we can check the dateIssued in PaymentEvent using PCSCase.dateIssued.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
```diff
@@ -95,4 +95,9 @@ public void patchCaseLinks(long caseReference, PCSCase pcsCase) {
             caseLinkService.mergeCaseLinks(pcsCase.getCaseLinks(), pcsCaseEntity);
         }
     }
+
+    public void setCaseIssuedDate(long caseReference) {
+        PcsCaseEntity pcsCaseEntity = loadCase(caseReference);
+        claimService.setClaimIssuedDate(pcsCaseEntity.getClaims().getFirst());
```
guygrewal77, line 117: Instead of passing the whole claimEntity, might be better to pass the claimId and let the ClaimService load the Claim using the id.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/ClaimService.java
```diff
@@ -88,4 +91,11 @@ private ClaimEntity buildClaimEntity(PCSCase pcsCase) {
             .build();
     }
 
+    public void setClaimIssuedDate(ClaimEntity claimEntity) {
```
guygrewal77: We can load the claimEntity using the id passed in the parameter instead. It's considered best practice to pass ids instead of whole entities.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseListView.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+
+    private void setDefendantNames(PCSCase pcsCase) {
+        List<ListValue<Party>> defendants = pcsCase.getAllDefendants();
+        if (CollectionUtils.isEmpty(defendants)) {
+            return;
+        }
+
+        String names = getDefendantName(defendants.getFirst().getValue());
```
guygrewal77: Could use Streams-based solution here to replace lines 42-46 for elegance.

    String names = defendants.stream()
            .limit(2)
            .map(ListValue::getValue)
            .map(this::getDefendantName)
            .collect(Collectors.joining(", "));

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseListView.java
```diff
@@ -0,0 +1,80 @@
...
+
+        if (defendants.size() > 1) {
+            names = names + ", " + getDefendantName(defendants.get(1).getValue());
+        }
+
+        if (defendants.size() > 2) {
+            names = names + " and Others";
```
guygrewal77: Could use shorthand operator: 

names += " and Others"

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseListView.java
gmmagruder: It shouldn't but thought it was a good idea to have a guard against npe exceptions just in case. Refactored this to be similar to how dateIssued is handled.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
gmmagruder: I disagree with this because we have to get the load the case entity anyway via case reference. We can pass the id through but we would already have the id at this point.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/ClaimService.java
gmmagruder: See reply from above

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseListView.java
gmmagruder: Addressed

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseListView.java
gmmagruder: Addressed

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/ClaimService.java
gmmagruder: Addressed

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/event/PaymentEvent.java
gmmagruder: Addressed

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseListView.java
guygrewal77: Perhaps an exception should be thrown if found to be null ? Similar to how an exception is thrown if no claims returned from the repository.

guygrewal77 (approved):

PR #1939, opened 2026-06-05
HDPI-7035: link case notes to pcs_claim table instead of claim table

gmmagruder:
### Jira link

See [HDPI-7035](https://tools.hmcts.net/jira/browse/HDPI-7035)

### Change description

- Remove one to many relationship between claim and case_note table
- Add one to many relationship between pcs_case and case_note table
- Amend claim_id in case_note table to be nullable - will be fully dropped in separate PR https://github.com/hmcts/pcs-api/pull/1940

### Testing done

  * [ ] Yes
  * [ ] No

### Checklist


toby-plunkett (approved):

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseNoteServiceTest.java
```diff
@@ -78,12 +78,12 @@ void shouldSaveNewCaseNoteWithNoPreexistingCaseNotes() {
...
-        verify(claimRepository).save(claimEntityCaptor.capture());
+        ArgumentCaptor<PcsCaseEntity> pcsCaseEntityCaptor = ArgumentCaptor.forClass(PcsCaseEntity.class);
+        verify(pcsCaseRepository).save(pcsCaseEntityCaptor.capture());
 
-        ClaimEntity persistedClaimEntity = claimEntityCaptor.getValue();
-        assertThat(persistedClaimEntity.getCaseNotes()).hasSize(1);
-        CaseNoteEntity persistedCaseNote = persistedClaimEntity.getCaseNotes().getFirst();
+        PcsCaseEntity persistedCaseEntity = pcsCaseEntityCaptor.getValue();
+        assertThat(persistedCaseEntity.getCaseNotes()).hasSize(1);
+        CaseNoteEntity persistedCaseNote = persistedCaseEntity.getCaseNotes().getFirst();
```
sadmanrahh, line 86: Can we assert that the persisted CaseNoteEntity is linked back to the PcsCaseEntity? That directly covers the new case-level relationship.

src/main/resources/db/migration/V120__link_case_note_table_to_pcs_case.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE case_note
+  ADD COLUMN case_id UUID REFERENCES pcs_case (id),
+  ALTER COLUMN claim_id DROP NOT NULL;
```
sadmanrahh, line 3: Should we backfill case_id from the existing claim_id relationship before dropping claim_id? I think case notes would have case_id = null and may no longer display after this refactor.

gmmagruder: I don't think this is necessary since we haven't gone live with the service yet. I believe the data in the lower environments gets deleted every day so I think we don't need to migrate the existing data in the lower environments over to the new approach though this may break any existing cases with case notes. Think the merge strategy with this is merge this PR first then merge the cleanup PR the next day.

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseNoteServiceTest.java
gmmagruder: Added

sadmanrahh (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseNoteView.java
```diff
@@ -5,19 +5,15 @@
...
-import uk.gov.hmcts.reform.pcs.ccd.entity.ClaimEntity;
 import uk.gov.hmcts.reform.pcs.ccd.entity.PcsCaseEntity;
 
 import java.util.List;
-import java.util.Optional;
 
 @Component
 public class CaseNoteView {
 
     public void setCaseFields(PCSCase pcsCase, PcsCaseEntity pcsCaseEntity) {
```
guygrewal77, line 15: We shouldn't have entities in the presentation layer at all. It's best to use a service to retrieve the data from the DB and return back the domain object.

gmmagruder: This is what we're currently doing for the other views? I think we need to do a lot of refactoring if you want to do it this way.

guygrewal77: I think the refactoring can be revisited later. Also the script from the other PR can be combined in this PR.

gmmagruder: The script from the other PR can't be combined into this one because it will cause issues when merging. This happens when dropping columns or tables.

guygrewal77: What issues does it cause ? I was able to run one after the other with case notes data locally.

gmmagruder: It breaks the functional tests in the master pipeline because it expects the column or table to be there but has been dropped.

guygrewal77 (approved):

PR #1941, opened 2026-06-07
HDPI-4276: Add wiremock stubs for more payment-api calls

scottstewart-sl:
### Jira link

See [HDPI-4276](https://tools.hmcts.net/jira/browse/HDPI-4276)

### Change description

Add local wiremock stubs for payment API:

* Create payment request: POST /payments/service-request/{serviceRequestReference}/card-payments
* Get status of payment request, (with internal payment reference): GET /payments/card-payments/{internalReference}/status
* Basic web page to simulate Gov Pay pages: GET /payment-web/card-payment

This PR also changes the generated service request ID on the existing stubbing to use the unique wiremock request ID rather than deriving it from the case reference, as it needs to be unique for all service requests on a given case.

Screenshot of fake payment web:

fake-payment-web

Selecting one of the options redirects the browser to the calling application with a fixed internal payment ref that corresponds to the chosen status

### Testing done

Tested locally with upcoming changes to pcs-frontend this ticket HDPI-4276

  * [ ] Yes
  * [x] No

### Checklist


wiremock/mappings/payment-api/payment-success.json
```diff
@@ -12,7 +12,7 @@
       "Content-Type": "application/json"
     },
     "jsonBody": {
-      "service_request_reference": "SR-{{{jsonPath request.body '$.ccd_case_number'}}}"
+      "service_request_reference": "SR-{{request.id}}"
```
tvr-hmcts, line 15: Interesting call as this should marry up with the original from the service request itself.   Does this still work with the validation?

scottstewart-sl: Could you clarify which validation you are referring to? This is the response to the create service request call, so is the first time the service request reference is generated.

Under normal operation, the service request reference would be generated by the payment API and could be any (unique) value. We don't perform any validation on it as far as I know.

arun-hmcts (approved):

libanAbdirahman1 (approved):

fudge88 (approved):

wiremock/mappings/payment-api/payment-success.json
scottstewart-sl: As discussed on the chat, I've tested this in ExUI locally and confirmed that the callback correlates to the expected service request entry in fee_payment, and cause the case state to be updated to ISSUED

tvr-hmcts (approved):

PR #1944, opened 2026-06-08
HDPI-5486-Add-column-(region_id)-to-pcs_case-table

tvr-hmcts:
https://tools.hmcts.net/jira/browse/HDPI-5486

* Added a region_id to the pcs_case
* Changed the way the case management location is held and when it is held.
* Added the allocation of the court venue to the pcsCase


src/main/java/uk/gov/hmcts/reform/pcs/location/service/LocationReferenceService.java
```diff
@@ -23,16 +24,17 @@ public class LocationReferenceService {
...
-    public List<CourtVenue> getCountyCourts(String authorisation, List<Integer> epimIds) {
+    public List<CourtVenue> getCourtVenues(List<Integer> epimIds) {
         if (Objects.isNull(epimIds) || epimIds.isEmpty()) {
             throw new IllegalArgumentException("epimIds cannot be null or empty");
         }
         String formattedEpimIds = formatEpimIds(epimIds);
         log.debug("Getting County courts from /refdata/location/court-venues for EpimIds {}", formattedEpimIds);
-        return locationReferenceApi.getCountyCourts(authorisation, authTokenGenerator.generate(),
-                formattedEpimIds, countyCourtTypeId);
+        return locationReferenceApi.getCourtVenues(authTokenGenerator.generate(),
```
sadmanrahh: The token order looks swapped here. LocationReferenceApi expects the IDAM token as Authorization and the S2S token as ServiceAuthorization, but this passes them the other way around. Can we swap these and update the test to verify the correct order?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/makeaclaim/StatementOfTruth.java
```diff
@@ -42,4 +51,13 @@ public void addTo(PageBuilder pageBuilder) {
             .done()
             .label("statementOfTruth-saveAndReturn", CommonPageContent.SAVE_AND_RETURN);
     }
+
+    private AboutToStartOrSubmitResponse<PCSCase, State> midEvent(CaseDetails<PCSCase, State> details,
+                                                                  CaseDetails<PCSCase, State> detailsBefore) {
+        if (allocateRegionId) {
+            pcsCaseService.allocateRegionId(details.getData());
+        }
+        return AboutToStartOrSubmitResponse.<PCSCase, State>builder().build();
```
sadmanrahh: This mid-event mutates the case data but returns no data in the response. Can we return .data(details.getData()) so the allocated regionId/caseManagementLocationNumber is written back to CCD before submission?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
```diff
@@ -90,13 +93,20 @@ public PcsCaseEntity loadCase(long caseReference) {
...
+        pcsCase.setCaseManagementLocationNumber(caseManagementLocation);
+    }
+
+    public void allocateRegionId(PCSCase pcsCase) {
+        allocateCaseManagementLocation(pcsCase);
+        if (pcsCase.getCaseManagementLocationNumber() != null) {
+            List<CourtVenue> courtVenues = locationReferenceService
+                .getCourtVenues(List.of(pcsCase.getCaseManagementLocationNumber()));
+            if (courtVenues != null && !courtVenues.isEmpty()) {
+                pcsCase.setRegionId(courtVenues.getFirst().epimmsId());
```
sadmanrahh: This appears to save the court EPIMS ID as regionId. Should this use the actual region_id from location-reference instead, since CaseLocation.region expects the region value rather than the base location?

src/main/java/uk/gov/hmcts/reform/pcs/postcodecourt/service/PostCodeCourtService.java
```diff
@@ -29,18 +29,16 @@ public PostCodeCourtService(PostCodeCourtRepository postCodeCourtRepository,
...
         List<PostCodeCourtEntity> results = getPostcodeCourtMappings(postCode);
-
         if (results.isEmpty()) {
-            log.error("EpimId not found, Case management location couldn't be allocated for postcode: {}", postCode);
+            log.error("EpimId not found, Court management location couldn't be allocated for postcode: {}", postCode);
             return null;
         }
         if (results.size() > 1) {
-            log.error("Multiple EpimId's found, Case management location not allocated for postcode: {}", postCode);
+            log.error("Multiple EpimId's found, Court management location not allocated for postcode: {}", postCode);
```
sadmanrahh: Minor - maybe should be 'EpimIds' instead

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseServiceTest.java
```diff
@@ -93,7 +97,8 @@ void setUp() {
             addressMapper,
             caseLinkService,
             caseFlagService,
-            postCodeCourtService
+            postCodeCourtService,
+            locationReferenceService
         );
     }
```
sadmanrahh, line 104: Can we add a test for when location-reference returns an empty list? That would confirm regionId remains null and the submit flow does not fail unexpectedly.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
```diff
@@ -60,7 +62,8 @@ public void createMainClaimOnCase(long caseReference, PCSCase pcsCase) {
         pcsCaseEntity.addDocuments(documentEntities);
         claimEntity.addClaimDocuments(documentEntities);
         pcsCaseEntity.addClaim(claimEntity);
-
+        pcsCaseEntity.setCaseManagementLocation(pcsCase.getCaseManagementLocationNumber());
```
benouaer, line 65: is it possible to rename this as epimsId or baseLocation, as caseManagement location is an object that includes both region and baseLocation

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
tvr-hmcts: Ah thank you.  Caught up and missed the obvious.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
tvr-hmcts: I found this quite confusing also TBH  I can rename.  BaseLocation?   There appears to be bot epims_id and epimms_id used within the api - which is also confusing.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/makeaclaim/StatementOfTruth.java
```diff
@@ -42,4 +51,15 @@ public void addTo(PageBuilder pageBuilder) {
             .done()
             .label("statementOfTruth-saveAndReturn", CommonPageContent.SAVE_AND_RETURN);
     }
+
+    private AboutToStartOrSubmitResponse<PCSCase, State> midEvent(CaseDetails<PCSCase, State> details,
+                                                                  CaseDetails<PCSCase, State> detailsBefore) {
+        if (allocateRegionId) {
+            pcsCaseService.allocateRegionId(details.getData());
```
benouaer, line 58: can we set the caseManagementLocation here, i.e. both region and baseLocation, because I think CCD will check that rather than regionId to grant access to the case

tvr-hmcts:
There is a problem with the pipeline workspace.

Closing this so to open a new PR (easier than fixing the workspace it appears as I do not have permission to do so) ... then will reference back to here from the new one.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/makeaclaim/StatementOfTruth.java
tvr-hmcts: Ah it does.  allocatedRegionId calls that internally and sets it on the PCSCase also.

PR #1945, opened 2026-06-08
Hdpi 7026 without notice doc upload flow for dashboard

fudge88:
### Jira link

See [HDPI-7026](https://tools.hmcts.net/jira/browse/HDPI-7026)

### Change description

Before, the uploadDocuments start handler in the API listed every submitted geApp on the case as a radio option on the "do these documents relate to an application?" page — regardless of whether the application was with notice or without notice. That meant a defendant could see that the claimant had filed a without-notice application.

Now, before building that list, the handler runs each genApp through the existing GenAppVisibilityService. The rule the service already enforces — and that we now rely on here — is: with-notice apps are visible to everyone, without-notice apps are visible only to the applicant who made them or their active legal representative. Anyone else gets a filtered-out list.

No frontend changes needed.

### Testing done
defendant 1 - without notice
can see their own genApp:
image

defendant 1 - without and with notice
can see both their apps:
image

defendant 2 - upload additional documents
can be the genApp that defendant 1 gave with notice:
image

  * [ ] Yes
  * [x] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/citizen/UploadDocuments.java
```diff
@@ -70,8 +72,10 @@ private PCSCase start(EventPayload<PCSCase, State> eventPayload) {
         }
 
         PcsCaseEntity pcsCaseEntity = pcsCaseService.loadCase(caseReference);
+        UUID currentUserId = securityContextService.getCurrentUserId();
 
         List<ListValue<RelatedApplicationOption>> options = visibleGenApps(pcsCaseEntity)
+            .filter(genApp -> genAppVisibilityService.isGenAppVisibleToUser(genApp, currentUserId))
```
scottstewart-sl: This `genAppVisibilityService` is (soon) going to filter out gen apps that haven't been issued yet. Is there a requirement for the user to be able to upload documents to gen apps that have been submitted and not issued yet? (e.g. waiting for payment).

In any case, it feels as if `visibleGenApps()` and `genAppVisibilityService.isGenAppVisibleToUser()` here are doing similar things and perhaps that should only be in one place.

toby-plunkett: should be perhaps do this filter inside the visibleGenApps call? I see that we handle all the other filters in there too. (I guess this depends on if visibleGenApps is used elsewhere or we are keeping this specific filter separate for some specific reason)

fudge88: yes you are right - i've asked the BAs just waiting for them to get back to me on this.

fudge88: it would be cleaner - im waiting on the BAs to get back to me on Scotts comment and will address this

fudge88: - BA confirmed only paid/issued genApps should be visible.
- visibleGenApps is now removed. start and submit now both route through genAppVisibilityService.getVisibleGenAppsToUser().

fudge88: - removed visibleGenApps entirely
- all filtering now lives in GenAppVisibilityService.getVisibleGenAppsToUser

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/citizen/UploadDocuments.java
```diff
@@ -87,14 +80,11 @@ private PCSCase start(EventPayload<PCSCase, State> eventPayload) {
         return caseData;
     }
 
-    private Stream<GenAppEntity> visibleGenApps(PcsCaseEntity pcsCaseEntity) {
-        if (pcsCaseEntity == null || pcsCaseEntity.getGenApps() == null) {
-            return Stream.empty();
+    private List<GenAppEntity> visibleGenAppsForUser(PcsCaseEntity pcsCaseEntity, UUID currentUserId) {
+        if (pcsCaseEntity == null) {
+            return List.of();
```
scottstewart-sl: Minor - `pcsCaseEntity` won't be null because you've just read it from the DB, (and it it didn't exist then an exception would be thrown)

scottstewart-sl (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/citizen/UploadDocuments.java
fudge88: removed the null check

toby-plunkett (approved):

PR #1952, opened 2026-06-08
HDPI-6034: Schedule a payment service request for gen apps in ExUI

scottstewart-sl:
### Jira link

See [HDPI-6034](https://tools.hmcts.net/jira/browse/HDPI-6034)

### Change description

Schedule a payment service request for gen apps in ExUI for applications that have a fee.

Other changes:
* Removing the `scheduleFeeCaseIssuedInSeconds` from the SchedulerConfig as hopefully it isn't necessary. 
* Adding a unique constraint on the service_request_reference in `fee_payment` as these values are expected to be unique, (and the code throws an exception in the callback method if they are not).
* Setting the payment responsible party name on the `FeesAndPayTaskData` because we don't always have a case `PartyEntity` corresponding to the payer, (e.g. it could be a legal rep, like in ExUI).

### Testing done

Manual testing in preview environment.

  * [ ] Yes
  * [x] No

### Checklist


arun-hmcts (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/genapp/SubmitEventHandler.java
```diff
@@ -131,22 +139,38 @@ private SubmitResponse<State> handleCuiSubmit(boolean paymentRequired,
...
 
-    private String createPaymentServiceRequest(GenAppEntity genAppEntity, FeeDetails feeDetails, long caseReference) {
-        UUID currentUserId = securityContextService.getCurrentUserId();
-        PartyEntity responsibleParty = partyService.getPartyEntityByIdamId(currentUserId, caseReference);
+    private void schedulePaymentServiceRequest(GenAppEntity genAppEntity, long caseReference, FeeDetails feeDetails) {
+        FeesAndPayTaskData taskData = createFeesAndPayTaskData(genAppEntity, caseReference, feeDetails);
+
+        schedulerClient.scheduleIfNotExists(
+            FEES_AND_PAY_TASK_DESCRIPTOR
+                .instance(UUID.randomUUID().toString())
```
gergelykisshmcts, line 151: Because the id is random, if the CCD submit callback is retried, the same general application could end up with multiple scheduled service-request creation tasks. Worth raising/flagging with BA even if we leave it like this and solve this issue manually by support latter.

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentService.java
```diff
@@ -86,7 +81,7 @@ public PaymentServiceResponse createServiceRequest(FeesAndPayTaskData feesAndPay
         FeeDto feeDto = paymentRequestMapper.toFeeDto(feesAndPayTaskData.getFeeDetails(),
                                                       feesAndPayTaskData.getVolume());
         CasePaymentRequestDto casePaymentRequest = paymentRequestMapper.toCasePaymentRequest(
-            getResponsiblePartyName(feesAndPayTaskData));
+            feesAndPayTaskData.getResponsiblePartyName());
         log.info("casePaymentRequest: {}", casePaymentRequest);
```
gergelykisshmcts: ResponsiblePartyName: this log writes the applicant name to application logs because it logs the whole object. Could we reduce this to non-personal identifiers?

src/test/java/uk/gov/hmcts/reform/pcs/ccd/event/genapp/SubmitEventHandlerTest.java
```diff
@@ -139,6 +155,57 @@ void shouldCreateGenAppWithCaseDataAndRepresentedParty() {
...
+            // When
+            underTest.submit(eventPayload(caseData));
+
+            // Then
+
+            if (feeApplies) {
+                verify(schedulerClient).scheduleIfNotExists(schedulableInstanceCaptor.capture());
+                SchedulableInstance<FeesAndPayTaskData> schedulableInstance = schedulableInstanceCaptor.getValue();
+                FeesAndPayTaskData feesAndPayTaskData = schedulableInstance.getTaskInstance().getData();
+
```
gergelykisshmcts, line 203: Can we assert if contains the exact fee amount this test currently would pass if the wrong fee detail was scheduled?

src/main/resources/db/migration/V127__fee_payment_unique_service_ref.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE fee_payment
+  ADD CONSTRAINT unique_service_request_ref UNIQUE(service_request_reference)
```
gergelykisshmcts, line 2: Is it confirmed there are no existing duplicates in the database? If not could we add a pre-deploy check for existing duplicate before this constraint is applied otherwise if any environment already contains duplicates, this migration will fail at startup.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/genapp/SubmitEventHandler.java
scottstewart-sl: It's a good point, but these scheduled tasks are written to the DB in the same transaction as the CCD submit handler, so if the submit fails they would be rolled back. And if the submit works, the CCD event has an idempotency token on it to effectively ignore a retry of the same submit. If the user resubmits the page on the frontend, (thereby creating a new event invocation), then we already have a check for duplicate submissions:

https://github.com/hmcts/pcs-api/blob/master/src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/genapp/SubmitEventHandler.java#L54-L56

src/main/java/uk/gov/hmcts/reform/pcs/feesandpay/service/PaymentService.java
scottstewart-sl: Yes, good point. That was existing code and probably left over from when the payments integration was being first tested in AAT. I've removed the log line completely now.

src/test/java/uk/gov/hmcts/reform/pcs/ccd/event/genapp/SubmitEventHandlerTest.java
scottstewart-sl: I've added an assertion for that.

src/main/resources/db/migration/V127__fee_payment_unique_service_ref.sql
scottstewart-sl: I have double checked the deployed environments, (AAT,Demo etc), and there are none.

The duplicate service request references were arising because the wiremock stubbing for creating a service request was generating the reference based on the CCD case reference in the request. That was fine when there was only 1 service request for a case, (i.e. pay for the claim), but soon we will have gen app payments as well. However, that code is still only in one of my branches, so no-one else should have duplicates and I've already updated the wiremock stubbing to create a unique service reference from now on: https://github.com/hmcts/pcs-api/pull/1941

gergelykisshmcts (approved):
All raised comments addressed. Looks good

gergelykisshmcts (approved):

tvr-hmcts (approved):
LGTM

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/genapp/SubmitEventHandler.java
```diff
@@ -135,22 +143,38 @@ private SubmitResponse<State> handleCuiSubmit(boolean paymentRequired,
...
```
saurabhvalsangkar-hmcts, line 151: Can we use something like GEN_APP_PAYMENT_ + genAppEntity.getId() to avoid possibility of duplicate task ?

PR #1984, opened 2026-06-16
HDPI-5819 Pin pack generation on case issue

gergelykisshmcts:
### Jira link

See [HDPI-5819](https://tools.hmcts.net/jira/browse/HDPI-5819)

### Change description

Generates the defendant access code letter (pin pack) when a case is issued, so each
defendant has a printable letter containing a unique access code they can use to respond
online or by post.

### Testing done

Added unit tests for the new code paths and updated tests for the modified ones.

Local build verification:

gradlew build end-to-end on this branch is green:

The test-pin sink was deliberately refactored from a JPA entity + repository to a plain
JdbcTemplate writer + reader. This keeps the table out of Hibernate's schema validation, so
the @DataJpaTest slice and other JPA-validating tests can boot without the runtime-created
table being present.

### Manual end-to-end (local stack)

Verified live by issuing real cases through XUI / Make a claim against bootWithCCD,
docmosis-tornado, and the local wiremock stubs for doc-assembly + Location Reference. 

 Case ref Scenario verified  :
 **1781600966587792** AC01 : 1 v 1 (John Smith) 1× DEFENDANT_ACCESS_CODE doc with category_id NULL; 1 BCrypt row in party_access_code; 1 SUCCESS log row; 1 plaintext row in testing_support_pin; GET /testing-support/pins/{ref} returns one entry; rendered PDF shows the right pin + name + addresses + court block.  
 **1781606860693552** AC02 : 1 v many (John Smith + Jane Doe) 2 docs (category NULL each), 2 **distinct** BCrypt codes, 2 SUCCESS log rows, 2 **distinct** plaintext pins (WDAHVW5MS7MV, EYP2MF8REKBP); /pins JSON has two top-level keys; one PDF per defendant.  
 **1781614122802623** AC04 + AC05 : persons unknown + no-address (one of each) DB shape identical to AC02; /pins returns {} for the persons-unknown / no-address defendant and an entry **without an address key** for the named / no-address defendant : the two response shapes downstream uses to detect each fallback; rendered PDFs show "Persons unknown" and property-address fallback.  
 All three AC03 : BCrypt at rest party_access_code.code starts with $2a$10$ (BCrypt), distinct per defendant, different from the plaintext value in testing_support_pin.plaintext_code. Plaintext never lives in party_access_code.  
 All three AC06 : respond-by-post court The court block on each rendered letter shows the venue returned by rd-location-ref-api (court_name, court_address, postcode), verified against the canonical LrdCourtVenueResponse schema in hmcts/rd-location-ref-api.  
 All three Hidden from Case File document.category_id is NULL for every DEFENDANT_ACCESS_CODE row (the DocumentService.mapDocumentTypeToCategory switch returns Optional.empty() for this type).  
 All three Failure logging Forced a failure by killing the location-reference stub mid-run: claim_activity_log recorded DOCUMENTS_CREATED / FAILURE in a new transaction (visible despite the outer rollback); db-scheduler retried with backoff; subsequent attempt logged SUCCESS.  

Evidence rendered locally (one PDF per defendant per case, via Docmosis Tornado direct :
same engine pcs-api hits via doc-assembly in prod):
- case-1781600966587792-defendant1-9LAJ7HBXTHR6.pdf
- case-1781606860693552-defendant1-WDAHVW5MS7MV.pdf
- case-1781606860693552-defendant2-EYP2MF8REKBP.pdf
- case-1781614122802623-defendant1-B9KGP5CXMNXN.pdf (persons unknown, property-address fallback)
- case-1781614122802623-defendant2-WX4598X7HBKS.pdf (no own address, property-address fallback)

Each was field-checked for: plaintext pin, defendant name (or
"Persons unknown"), case reference, property address, court address + postcode, the
"Respond by: The date of the hearing" label, and the /access-your-case URL.

  * [ ] Yes
  * [x] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/testingsupport/config/TestingSupportSchemaInitializer.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+@Component
+@RequiredArgsConstructor
+@Slf4j
+public class TestingSupportSchemaInitializer {
+
+    private final JdbcTemplate jdbcTemplate;
+
+    @Value("${pin-test-table.enabled:false}")
+    private boolean enabled;
```
scottstewart-sl: Minor - could you make this whole Component conditional on that flag, so (a) it wouldn't be created at all in Prod, and (b) it matches the approach used in `NoOpTestPinRecorder` ?

src/main/resources/application.yaml
```diff
@@ -162,7 +162,13 @@ notify:
     general-application-received: e5daba0d-5c66-4c76-9e0f-06646200dc89
 
 testing-support:
-  enabled: ${ENABLE_TESTING_SUPPORT:true}
+  enabled: ${ENABLE_TESTING_SUPPORT:false}
+
+pin-pack:
+  respond-online-url: ${RESPOND_ONLINE_URL:https://pcs.aat.platform.hmcts.net/access-your-case}
```
scottstewart-sl: I think it is better to make the default be a localhost URL. That way if the override isn't set correctly in the values.yaml then it will fail on AAT, rather than only failing later on when testing is done on an different environment like Demo.

On that note, do you need to add `RESPOND_ONLINE_URL` to the values.yaml as part of this PR, or will that go into the Flux config?

src/main/java/uk/gov/hmcts/reform/pcs/testingsupport/service/PersistingTestPinRecorder.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.util.UUID;
+
+/**
+ * Non-production recorder that persists plaintext pins to {@code testing_support_pin} for the QA pins endpoint.
+ */
+@Service
+@RequiredArgsConstructor
+@Slf4j
+@ConditionalOnProperty(name = "pin-test-table.enabled", havingValue = "true", matchIfMissing = true)
```
scottstewart-sl: Minor - I think it is safer to make the NoOpTestPinRecorder the default if the config property is missing, rather than this one.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/accesscode/AccessCodeFormDocumentGenerator.java
```diff
@@ -0,0 +1,170 @@
...
+        if (defendant.getNameKnown() == VerticalYesNo.YES) {
+            String name = formatPartyName(defendant);
+            if (StringUtils.isNotBlank(name)) {
+                return name;
+            }
+        }
+        return PERSONS_UNKNOWN;
+    }
+
+    private String formatPartyName(PartyEntity party) {
```
scottstewart-sl, line 112: Btw, `PartyService` has a method `getPartyName` that does something similar to this. Maybe we should consolidate that and just have a single place for getting a party name

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/accesscode/AccessCodeFormDocumentGenerator.java
```diff
@@ -0,0 +1,170 @@
...
+            && defendant.getAddressSameAsProperty() != VerticalYesNo.YES
+            && defendant.getAddress() != null) {
+            return formatAddress(defendant.getAddress());
+        }
+        return formattedPropertyAddress;
+    }
+
+    private CourtVenue resolveServingCourt(PcsCaseEntity pcsCaseEntity) {
+        Integer epimsId = pcsCaseEntity.getCaseManagementLocation();
+        if (epimsId == null) {
```
sadmanrahh, line 133: If caseManagementLocation is missing, we still generate the pin pack with null court details. AC06 requires the respond-by-post court address, so should this fail and retry rather than create a successful incomplete letter?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/AccessCodeGenerationService.java
```diff
@@ -50,22 +54,62 @@ public void createAccessCodesForParties(String caseReference) {
...
 
-        List<PartyEntity> defendants = getMainClaimDefendants(pcsCaseEntity);
+        ClaimEntity mainClaim = getMainClaim(pcsCaseEntity);
 
-        Set<PartyAccessCodeEntity> newCodes = defendants.stream()
-            .filter(d -> !existingPartyIds.contains(d.getId()))
-            .map(d -> createPartyAccessCodeEntity(pcsCaseEntity, d.getId()))
-            .collect(Collectors.toSet());
+        List<PartyEntity> newDefendants = getMainClaimDefendants(mainClaim).stream()
+            .filter(defendant -> !existingPartyIds.contains(defendant.getId()))
```
sadmanrahh: This filter skips any defendant that already has an access-code row. If the code exists but the DEFENDANT_ACCESS_CODE document is missing, we will never generate/repair the pin-pack document for that defendant. Can the idempotency check include the document as well as the code?

src/main/resources/application.yaml
```diff
@@ -232,7 +238,7 @@ fees:
     backoff-delay-seconds: ${FEES_AND_PAY_TASK_BACKOFF_DELAY_SECONDS:600s}
 
 access-code:
-  hash-pins-enabled: ${HASH_PINS_ENABLED:false}
+  hash-pins-enabled: ${HASH_PINS_ENABLED:true}
```
sadmanrahh: Now that the plaintext implementation has been removed and HashingPartyAccessCodeService is unconditional, is this property still needed? Should we remove this config and the related HASH_PINS_ENABLED chart values?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/AccessCodeGenerationService.java
gergelykisshmcts: Good thought, but the current filter is sufficient: createAccessCodesForParties is @Transactional and generateAccessCodeAndPinPack saves the document, the access code, the plaintext, and the success log all inside the same transaction with no intermediate commits, so we can't reach a state where the code is persisted but the document isn't. Any exception in the save path rolls all four back, the scheduled task retries with exponential backoff, and the defendant is picked up again because there's no party_access_code row. The check would only be insufficient if a document row was deleted out of band.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/accesscode/AccessCodeFormDocumentGenerator.java
gergelykisshmcts: Good spot: the two methods do look very similar at first glance. The reason I kept a separate one in PinPackDocumentGenerator rather than reusing PartyService.getPartyName is the null-handling: PartyService.getPartyName returns firstName + " " + lastName directly, which produces strings like "null Smith", "John null" or a leading, trailing space when only one of the name parts is set, and never returns null for an entirely empty party. Our resolveDefendantName relies on getting back null/empty for an empty party so it can fall back to "Persons unknown" (AC04), and it needs the filtered/trimmed join so partial-name defendants render cleanly on the letter.
I did consider just extending PartyService.getPartyName to be null-safe, but it's already used elsewhere (like PaymentService) and changing its return semantics (introducing null returns and dropping the always-present space) felt like it could have knock-on effects in callers I haven't fully traced for this ticket and not in the scope of refactoring that.

src/main/java/uk/gov/hmcts/reform/pcs/testingsupport/config/TestingSupportSchemaInitializer.java
gergelykisshmcts: done

src/main/resources/application.yaml
gergelykisshmcts: done

src/main/java/uk/gov/hmcts/reform/pcs/testingsupport/service/PersistingTestPinRecorder.java
gergelykisshmcts: done

src/main/resources/application.yaml
gergelykisshmcts: done

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/accesscode/AccessCodeFormDocumentGenerator.java
gergelykisshmcts: done

sadmanrahh (approved):

scottstewart-sl (approved):

PR #1985, opened 2026-06-17
HDPI-7209: Support multiple fee payments per claim

andrewdami11:
### Jira link

[HDPI-7209](https://tools.hmcts.net/jira/browse/HDPI-7209)

Infrastructure prerequisite for [HDPI-5806](https://tools.hmcts.net/jira/browse/HDPI-5806) (counterclaim fee payment). Extracted from the 5806 BE branch so the data-model fix can land on `master` before the feature PR.

**Branch:** `fix/claim-fee-payment-one-to-many` → `master`

### Change description

A claim can have more than one fee payment (e.g. initial claim issue fee and a later counterclaim fee). The JPA mapping incorrectly treated this as `@OneToOne`, which breaks when a second payment is saved for the same claim.

This PR:

- Changes `ClaimEntity` to hold `feePayments` as `@OneToMany` and `FeePaymentEntity.claim` as `@ManyToOne` (no DB migration needed, the table already allows multiple rows per claim).
- Persists the responsible `party` when saving a new fee payment via `PaymentService.saveNewFeePayment`, using the existing `responsiblePartyId` on task data.
- Updates counterclaim payment notification lookup to find the correct payment from the `feePayments` list by matching party (replacing the previous single `feePayment` accessor).

### Testing done

Updated unit tests cover the one-to-many mapping, party persistence on save, and party-based fee payment lookup in notification services. The payment callback integration test setup was fixed so `responsiblePartyId` is assigned from the test case party.

### Security Vulnerability Assessment

- [ ] Yes
- [x] No

### Checklist


src/test/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationPersonalisationFactoryTest.java
```diff
@@ -302,8 +304,9 @@ void shouldIncludePaymentReferenceNumber() {
             FeePaymentEntity feePayment = FeePaymentEntity.builder()
                 .paymentStatus(PaymentStatus.PAID)
                 .externalReference("PAY-123")
+                .party(defendantParty)
                 .build();
-            response.getClaim().setFeePayment(feePayment);
+            response.getClaim().setFeePayments(new ArrayList<>(List.of(feePayment)));
```
scottstewart-sl: Minor - out of interest why does this need to be converted to an ArrayList before setting it?

Also generally when adding a child entity we have a custom "add" method that sets the parent on the child before adding it, for example:

https://github.com/hmcts/pcs-api/blob/master/src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/PcsCaseEntity.java#L139-L142

That would also make this code slightly simpler as you could just call `response.getClaim().addFeePayment(feePayment);`

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationPersonalisationFactory.java
```diff
@@ -64,18 +65,27 @@ public BasePersonalisation forParty(PartyEntity partyEntity, PcsCaseEntity pcsCa
...
             .base(forDefendant(defendantResponse))
             .paymentReferenceNumber(defendantFeePayment.getExternalReference())
             .build();
     }
 
+    private FeePaymentEntity findPaidFeePaymentForParty(ClaimEntity claim, UUID partyId, UUID defendantResponseId) {
+        return claim.getFeePayments().stream()
+            .filter(feePayment -> feePayment.getParty() != null && partyId.equals(feePayment.getParty().getId()))
+            .filter(feePayment -> PaymentStatus.PAID.equals(feePayment.getPaymentStatus()))
+            .findFirst()
```
scottstewart-sl: I don't think this is safe as it assumes that a party would only ever pay for a maximum of 1 fee. If they had two, and only the first one was paid, if they raised a second counterclaim this code would treated it as PAID wouldn't it?

src/test/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationPersonalisationFactoryTest.java
andrewdami11: The ArrayList wrapper was a bit of a shortcut to make the test work, but I've applied your suggestion now [here](https://github.com/hmcts/pcs-api/pull/1985/commits/57904d3a783bc23f9f92b6c88812baed9ab7fd91)

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationPersonalisationFactory.java
andrewdami11: Now addressed [here](https://github.com/hmcts/pcs-api/pull/1985/commits/a9b612053d157a001b2284a0a147e4e9474d55ab)

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationPersonalisationFactory.java
```diff
@@ -63,16 +63,18 @@ public BasePersonalisation forParty(PartyEntity partyEntity, PcsCaseEntity pcsCa
...
-    public CounterclaimPaymentSuccessPersonalisation counterclaimSuccess(DefendantResponseEntity defendantResponse) {
-        FeePaymentEntity defendantFeePayment = defendantResponse.getClaim().getFeePayment();
-        if (defendantFeePayment == null || !defendantFeePayment.getPaymentStatus().equals(PaymentStatus.PAID)) {
+    public CounterclaimPaymentSuccessPersonalisation counterclaimSuccess(
+        DefendantResponseEntity defendantResponse,
+        FeePaymentEntity feePayment
+    ) {
+        if (feePayment == null || !PaymentStatus.PAID.equals(feePayment.getPaymentStatus())) {
             throw new FeePaymentNotFoundException(
                 "Paid fee payment not found for defendant response: " + defendantResponse.getId());
```
scottstewart-sl, line 72: Minor - not part of your change, but you could mention in this log line that it is a counterclaim fee payment, (rather than just for the defendant response).

scottstewart-sl (approved):

arun-hmcts (approved):

PR #1994, opened 2026-06-18
HDPI-7091: remove duplicate documents in case file view

gmmagruder:
### Jira link

See [HDPI-7091](https://tools.hmcts.net/jira/browse/HDPI-7091)

### Change description

Remove documents that appear in case details tab from allDocuments so that they only appear once when viewing the case file view.

### Testing done

  * [ ] Yes
  * [ ] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/DocumentsView.java
```diff
@@ -63,8 +65,29 @@ public boolean isDocumentVisibleToUser(DocumentEntity documentEntity, UUID curre
...
+            DocumentType.OCCUPATION_LICENCE
+        );
+
+        DocumentType type = documentEntity.getType();
+        if (type == null || !caseDetailsDocuments.contains(type)) {
+            return true;
+        }
+
+        // Is not an additional document
+        return documentEntity.getDescription() != null;
```
guygrewal77: The static method - isDescriptionEmpty() can be used instead to check if it's an additional document.

gmmagruder: Addressed

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/DocumentsView.java
```diff
@@ -63,8 +65,29 @@ public boolean isDocumentVisibleToUser(DocumentEntity documentEntity, UUID curre
         }
     }
 
+    private boolean isNotInCaseDetailsTab(DocumentEntity documentEntity) {
+        List<DocumentType> caseDetailsDocuments = List.of(
```
guygrewal77, line 89: It's still possible to upload these documents in Upload Additional Documents page so it should be enough to differentiate between these documents uploaded and those from elsewhere from just checking the description field which will be mandatory.

gmmagruder: There may be some documents from other events (e.g. respond to a claim) which are not additional documents.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/DocumentsView.java
```diff
@@ -63,8 +65,29 @@ public boolean isDocumentVisibleToUser(DocumentEntity documentEntity, UUID curre
...
+            DocumentType.POSSESSION_NOTICE,
+            DocumentType.RENT_STATEMENT,
+            DocumentType.ENERGY_PERFORMANCE_CERTIFICATE,
+            DocumentType.EICR_REPORT,
+            DocumentType.GAS_SAFETY_CERTIFICATE,
+            DocumentType.OCCUPATION_LICENCE
+        );
+
+        DocumentType type = documentEntity.getType();
+        if (type == null || !caseDetailsDocuments.contains(type)) {
```
guygrewal77, line 100: Can a test be added to cover all scenarios in the if statement and scenarios for other journeys which don't set document types or descriptions ?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/builder/NoticeDetailsBuilder.java
```diff
@@ -71,17 +74,25 @@ private NoticeTabDetails buildNoticeTabDetailsWales(PCSCase pcsCase) {
...
+        NoticeServedDetails noticeServedDetails,
+        boolean isSubmitted
+    ) {
         if (noticeServedDetails == null || noticeServedDetails.getServiceMethod() == null) {
             return;
         }
 
-        noticeTabDetails.setNoticeDocuments(noticeServedDetails.getDocuments());
+        List<ListValue<Document>> documents = noticeServedDetails.getDocuments();
+        if (isSubmitted) {
```
guygrewal77, line 92: Is there a reason to be checking this ?

gmmagruder: Yes because PCSCaseView is also called on every page during events and we don't want to overwrite this data.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/DocumentsView.java
gmmagruder: Added

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseTabView.java
```diff
@@ -42,10 +42,10 @@ public class CaseTabView {
     private final CaseSummaryTabView caseSummaryTabView;
     private final CaseDetailsTabView caseDetailsTabView;
 
-    public void setCaseTabFields(PCSCase pcsCase) {
+    public void setCaseTabFields(PCSCase pcsCase, boolean isSubmitted) {
```
guygrewal77: Do we need this extra parameter ? PCSCase has dateSubmitted field which can be used to check if it's been submitted.

gmmagruder: Good point, have changed this to set isSubmitted within the method based on whether dateSubmitted is null or not.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseTabView.java
```diff
@@ -43,9 +43,10 @@ public class CaseTabView {
     private final CaseDetailsTabView caseDetailsTabView;
 
     public void setCaseTabFields(PCSCase pcsCase) {
+        boolean isSubmitted = pcsCase.getDateSubmitted() != null;
```
guygrewal77, line 46: We could put this check in a utility method which can be called from each of the builders, e.g. NoticeDetailsBuilder instead of passing this extra parameter. Each of the builders already has PcsCase passed to them.

gmmagruder: I believe it would be cleaner to pass it in as a parameter because it is only calculated once and the builders are only being called by the case details tab view. This will avoid unnecessary repeated calls and allows all builders to use the same result.

guygrewal77 (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseTabView.java
guygrewal77: What if a new requirement came along to check if the claim has been issued ? We could end up introducing more parameters.

toby-plunkett (approved):

PR #2002, opened 2026-06-19
Hdpi 3460 delete draft claim

sadmanrahh:
### Jira link

See [PROJ-3460](https://tools.hmcts.net/jira/browse/PROJ-3460)

### Change description

Allows the ability for users with (CREATOR role) to delete a draft possession claim (AWAITING_SUBMISSION_TO_HMCTS / PENDING_CASE_ISSUED) from the Next Steps dropdown. Selecting "Yes" on the confirmation screen hard-deletes all associated case data and removes it from the case list.

### Testing done

Screenshot 2026-06-19 at 15 17 25
Screenshot 2026-06-19 at 15 17 46
Screenshot 2026-06-19 at 15 18 02

  * [ ] Yes
  * [X] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/DeleteDraftClaim.java
```diff
@@ -0,0 +1,88 @@
...
+    public void configureDecentralised(DecentralisedConfigBuilder<PCSCase, State, UserRole> configBuilder) {
+        new PageBuilder(configBuilder
+                            .decentralisedEvent(DELETE_DRAFT_CLAIM.id(), this::submit)
+                            .forStates(State.AWAITING_SUBMISSION_TO_HMCTS, State.PENDING_CASE_ISSUED)
+                            .name("Delete draft claim")
+                            .showCondition(draftClaimStateCondition())
+                            .grant(Permission.CRUD, UserRole.CREATOR)
+                            .grant(Permission.CRUD, UserRole.CLAIMANT_SOLICITOR)
+                            .endButtonLabel("Continue"))
+            .page("deleteDraftClaim")
```
guygrewal77: I would prefer to take the approach taken for ConfirmEviction event which is also a relatively small event flow. That way we separate the page handling from the event configuration.

src/integrationTest/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftClaimDeletionServiceIT.java
```diff
@@ -0,0 +1,183 @@
...
+
+import java.sql.Timestamp;
+import java.time.Instant;
+import java.util.UUID;
+import java.util.concurrent.ThreadLocalRandom;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Testcontainers
+class DraftClaimDeletionServiceIT {
```
guygrewal77: I don't think we need an integration test here.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/DeleteDraftClaim.java
sadmanrahh: Sure, I have refactored to follow ConfirmEviction

src/integrationTest/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftClaimDeletionServiceIT.java
sadmanrahh: I added it to cover the deletion flow end-to-end, including persistence behavior. But agreed, existing coverage is enough here, so I’ve removed it.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CcdCaseDataService.java
```diff
@@ -0,0 +1,27 @@
...
+
+import lombok.AllArgsConstructor;
+import org.springframework.dao.EmptyResultDataAccessException;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.stereotype.Service;
+import uk.gov.hmcts.reform.pcs.ccd.domain.State;
+
+@Service
+@AllArgsConstructor
+public class CcdCaseDataService {
```
guygrewal77, line 11: Not sure about needing this class.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/EventId.java
```diff
@@ -16,5 +16,20 @@ public enum EventId {
     addCaseNote,
     createFlags,
     amendFlags,
-    claimIssuePayment
+    claimIssuePayment,
+    DELETE_DRAFT_CLAIM("deleteDraftClaim");
```
shazanm-soli: Change this to be just deleteDraftClaim. This keeps all the enum values in the same style

shazanm-soli:
After deleting a case, should it not go back to the list cases page? Returning back to the case overview, most of the tabs are empty and case file view displays an error message.

Screenshot 2026-06-23 at 10 33 59

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CcdCaseDataService.java
sadmanrahh: This was added to avoid a race where the post-create role assignment task can run after a draft has already been deleted. In that case we do not want to reassign CLAIMANTSOLICITOR to a deleted/missing CCD case. I can rename it or move the query closer to the role-assignment task if you think that would make the intent clearer.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/EventId.java
sadmanrahh: Sure, addressed

sadmanrahh:
Agreed, after delete AC05 says the draft should be removed from Case List entirely.
PCS deletes the data and removes it from Case List, but EXUI controls the post-submit navigation. It currently refetches/returns to the deleted case briefly; after a second or two the case is fully removed from Case List.
I’ve spoken to Stephen about this, and he advised noting it in the Jira as immediate removal/redirect from Case List would need an EXUI change.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftClaimDeletionService.java
```diff
@@ -0,0 +1,204 @@
...
+    public void deleteDraftClaim(long caseReference) {
+        UUID caseId = getCaseId(caseReference);
+
+        if (caseId == null) {
+            jdbcTemplate.update("DELETE FROM draft.draft_case_data WHERE case_reference = ?", caseReference);
+            deleteCcdCaseData(caseReference);
+            return;
+        }
+
+        if (hasIssuedClaim(caseId)) {
```
tvr-hmcts: Does it matter that this can change while this is running?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftClaimDeletionService.java
```diff
@@ -0,0 +1,204 @@
...
+            throw new IllegalStateException("Cannot delete a claim that has been issued");
+        }
+
+        jdbcTemplate.update("DELETE FROM draft.draft_case_data WHERE case_reference = ?", caseReference);
```
guygrewal77: Is there a reason to be using jdbcTemplate instead of jdbc repository ?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftClaimDeletionService.java
```diff
@@ -0,0 +1,204 @@
...
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+import java.util.List;
+import java.util.UUID;
+
+@Service
+@AllArgsConstructor
+public class DraftClaimDeletionService {
```
tvr-hmcts, line 13: Are any other approaches possible?  like through a single update or a cascade?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftClaimDeletionService.java
```diff
@@ -0,0 +1,204 @@
...
+        deleteRowsLinkedToCase(caseId);
+        deleteRowsLinkedToClaims(caseId);
+        deleteRowsLinkedToParties(caseId);
+
+        jdbcTemplate.update("DELETE FROM document WHERE case_id = ?", caseId);
+        jdbcTemplate.update("DELETE FROM claim WHERE case_id = ?", caseId);
+        jdbcTemplate.update("DELETE FROM party WHERE case_id = ?", caseId);
+        jdbcTemplate.update("DELETE FROM pcs_case WHERE id = ?", caseId);
+
+        contactPreferenceIds.forEach(contactPreferenceId ->
```
tvr-hmcts: These foreach can be done in a single batch update instead.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftClaimDeletionService.java
tvr-hmcts: Wondered this also - particularly with the first level cache in hibernate.   There maybe a need to flush and clear etc

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/DeleteDraftClaim.java
```diff
@@ -0,0 +1,88 @@
...
+        if (eventPayload.caseData().getDeleteDraftClaim() != YesOrNo.YES) {
+            return SubmitResponse.defaultResponse();
+        }
+
+        scheduleDraftClaimDeletion(eventPayload.caseReference());
+
+        return SubmitResponse.<State>builder()
+            .state(State.DELETED)
+            .confirmationBody("""
+                # Case deleted
```
tvr-hmcts, line 63: Should this say "Scheduled for deletion" as the delete itself could fail and it is not straight away.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/deletedraftclaim/DeleteDraftClaimConfigurer.java
```diff
@@ -0,0 +1,14 @@
...
+
+import org.springframework.stereotype.Component;
+import uk.gov.hmcts.reform.pcs.ccd.common.PageBuilder;
+import uk.gov.hmcts.reform.pcs.ccd.common.PageConfigurer;
+
+@Component
+public class DeleteDraftClaimConfigurer implements PageConfigurer {
+
+    @Override
+    public void configurePages(PageBuilder pageBuilder) {
```
tvr-hmcts, line 11: This maybe overkill for this single page unless there is more things to come?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/DeleteDraftClaimTaskComponent.java
```diff
@@ -0,0 +1,87 @@
...
+            ))
+            .execute((taskInstance, executionContext) -> {
+                DeleteDraftClaimTaskData taskData = taskInstance.getData();
+                long caseReference = Long.parseLong(taskData.getCaseReference());
+                String userId = taskData.getUserId();
+
+                log.debug("Deleting draft claim for case: {}", caseReference);
+
+                try {
+                    revokeCaseRoles(caseReference, userId);
```
tvr-hmcts, line 61: I think this should be after the delete as I do not think it will rollback like if the transaction fails.

tvr-hmcts (commented):
Some thoughts.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/deletedraftclaim/DeleteDraftClaimConfigurer.java
guygrewal77: It makes it consistent with other events and easier to test using existing patterns.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftClaimDeletionService.java
guygrewal77: Something like this ? 

@Entity
public class ChildEntity {
    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    private Long id;

    @ManyToOne
    @JoinColumn(name = "parent_id")
    // This tells Hibernate to generate DDL with "ON DELETE CASCADE"
    @OnDelete(action = OnDeleteAction.CASCADE) 
    private ParentEntity parent;

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftClaimDeletionService.java
sadmanrahh: Good point. I’ve moved the SQL into a dedicated JDBC repository so the service no longer depends on JdbcTemplate directly. I’ve avoided JPA repositories here because this is a hard delete across multiple tables/schemas and using JPA could introduce first-level cache/flush concerns.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftClaimDeletionService.java
sadmanrahh: Cascade delete could work as a longer-term data model approach, but I think it is too broad for this ticket. This hard delete spans draft, ccd, and multiple PCS tables, including some indirect/orphan cleanup. Adding cascade rules would require schema/entity changes and could affect other delete paths. I’ve kept this scoped with explicit deletes for this event only.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/DeleteDraftClaim.java
sadmanrahh: I kept this as “Case deleted” to match the AC/Figma confirmation wording, but if you want I can speak to Stephen regarding it

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftClaimDeletionService.java
sadmanrahh: Good point. I've added a row lock on claim rows before checking claim_issued_date, so a concurrent payment callback cannot issue the claim while the delete transaction is running.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/DeleteDraftClaimTaskComponent.java
sadmanrahh: We can't move the role revocation after the delete without breaking the case list removal. The CCD case assignment API needs the CCD case to still exist to remove the user's role, and deleteDraftClaim removes the CCD case at the end. So revoking after the delete would leave the role assignment in place and the case would stay in the case list. Keeping revokeCaseRoles before deleteDraftClaim ensures the case is removed from the case list first.

PR #2022, opened 2026-06-25
HDPI-6835: Gen App Doc Sequencing

sadmanrahh:
### Jira link

See [PROJ-6835](https://tools.hmcts.net/jira/browse/PROJ-6835)

### Change description

Move GA numbering from party level to case level:

-The first GA created on a case is GA1
-The second GA on the same case is GA2
-The third is GA3, and so on

This must apply regardless of:

-which party uploads the GA
-what type of general application it is

### Testing done

Screenshot 2026-06-25 at 15 04 44

  * [ ] Yes
  * [X] No

### Checklist


src/main/resources/db/migration/V128__gen_app_case_level_rank.sql
```diff
@@ -0,0 +1,19 @@
...
+-- Re-number any existing rows so that the first GA per case is rank 1, the second is rank 2, etc.
+WITH renumbered AS (
+    SELECT id,
+           ROW_NUMBER() OVER (PARTITION BY case_id ORDER BY application_submitted_date NULLS LAST, id) AS new_rank
+    FROM general_application
+)
+UPDATE general_application ga
+SET rank = r.new_rank
+FROM renumbered r
+WHERE ga.id = r.id;
```
scottstewart-sl, line 15: Nice to see this bit of the script to clean up the existing data, but actually some document filenames will have already been tagged with the existing GA numbers, so it would be more consistent to leave the existing data as it is, and accept that the numbering isn't correct for old data.

sadmanrahh: Good point, I've removed the data numbering from the migration so existing GA ranks and filenames already tagged with them remain unchanged. The case-level sequencing now applied to new GAs going forward

scottstewart-sl (approved):

andrewdami11 (approved):

gmmagruder (approved):

src/main/resources/db/migration/V128__gen_app_case_level_rank.sql
sadmanrahh: As discussed, I have re-added the renumbering step before applying the new (case_id, rank) constraint, so any existing duplicate case-level ranks are normalised before the constraint is created.

PR #2027, opened 2026-06-26
Reducing calls to Ref Data within the start up of Resume

tvr-hmcts:
[HDPI-5414](https://tools.hmcts.net/jira/browse/HDPI-5414)

Note - this is first run at this ticket.  As it has been highlighted further work is needed in other areas also from a performance perspective.

Highlighted in performance testing the start of the ResumePossessionClaim is not meeting SLAs.   Identified the double call to Ref Data and have changed the approach there.

I profiled this within an integration test and the results of the change are presented in the Jira.

Ran further hibernate based statistics and only one db call is made for the draft data.  We can consider additional work there if needed in the future as splitting this using projections.

Further large or more complex refactoring can/will be considered later also such as caching and parallel calls


src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/ResumePossessionClaim.java
```diff
@@ -90,20 +93,17 @@ public void configureDecentralised(DecentralisedConfigBuilder<PCSCase, State, Us
 
     }
 
-    private PCSCase start(EventPayload<PCSCase, State> eventPayload) {
+    PCSCase start(EventPayload<PCSCase, State> eventPayload) {
```
sadmanrahh, line 95: Why was this changed from private to package-private? The tests call the registered CCD handler, not this method directly.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/ResumePossessionClaim.java
```diff
@@ -136,7 +136,7 @@ private PCSCase start(EventPayload<PCSCase, State> eventPayload) {
             .build();
         caseData.setClaimantType(claimantTypeList);
 
-        contactPreferences.setOrganisationAddress(organisationService.getOrganisationAddressForCurrentUser());
+        contactPreferences.setOrganisationAddress(null != nameAndAddress ? nameAndAddress.address() : null);
```
sadmanrahh, line 138: Can we extract using AddressUK orgAddress = nameAndAddress != null ? nameAndAddress.address() : null; to avoid the second null check and make the flow clearer?

src/main/java/uk/gov/hmcts/reform/pcs/reference/service/OrganisationService.java
```diff
@@ -76,6 +91,8 @@ public String getOrganisationIdForCurrentUser() {
         }
     }
 
+
```
sadmanrahh, line 94: Minor- can we remove extra lines?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/ResumePossessionClaim.java
tvr-hmcts: Ah I placed an integration test against it at that point.  I can switch it back as Im not pushing the integration test as I only used it within the profiling.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/ResumePossessionClaim.java
tvr-hmcts: Based on the pattern in the OrganisationDetailsService it is possible for it to fail and then continue - returning null on exception.    Ive not changed that pattern, though perhaps that needs revisiting, so it is possible for the response (NameAndAddress) to be null.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/ResumePossessionClaim.java
sadmanrahh: Sure, if you could switch back, would be great!

src/main/java/uk/gov/hmcts/reform/pcs/reference/service/OrganisationService.java
```diff
@@ -23,6 +24,20 @@ public class OrganisationService {
     private final SecurityContextService securityContextService;
     private final OrganisationDetailsService organisationDetailsService;
 
+    public NameAndAddress getNameAndAddressForCurrentUser() {
+        try {
+            UUID userId = resolveUserId();
```
guygrewal77, line 29: There's some missing code coverage in a method implemented outside of this PR - getOrganisationAddressForCurrentUser()

tvr-hmcts: Though the rationale for this ticket is not really in this area per say but in how we use it.  I will take bit more of a look though now as a stepping stone.   I think this class and the OrganisationDetailsService may need a revisit in general and there is another ticket that may impact it also - though not in flight yet - https://tools.hmcts.net/jira/browse/HDPI-6650

guygrewal77 (approved):

sadmanrahh (approved):

PR #2094, opened 2026-06-27
HDPI-5947: HDPI-5947: Add LaunchDarkly feature toggle for bulk print

arun-hmcts:
### Jira link
HDPI-5947

### Change description
This PR adds LaunchDarkly to pcs-api so we can turn features on and off per environment without a redeploy.

It sets up the LaunchDarkly client, a FeatureToggleService to check flags, and a FeatureFlag enum to list them. The first flag is bulk-print-enabled, which will later control the bulk print task. Nothing uses it yet, so this change is safe to merge on its own — the value defaults to off until we switch it on in LaunchDarkly.

If LaunchDarkly can't be reached, the flag falls back to its default in code. Local and test runs use offline mode, and the real key comes from Key Vault in the deployed environments.

### Testing done

Tested on the PR preview pod — the LaunchDarkly client connects on startup and the bulk-print-enabled flag is read correctly for the environment. Flipped the flag in LaunchDarkly and confirmed the value changes live without a restart. Unit tests cover the flag registry, the toggle service, and the online/offline client setup.
And also restarted the pcs--api locally.

  * [] Yes
  * [x] No

### Checklist


build.gradle
```diff
@@ -550,6 +550,8 @@ tasks.withType(CftlibExec).configureEach {
   environment 'CCD_DECENTRALISED_CASE-TYPE-SERVICE-URLS_PCS', 'http://localhost:3206'
   environment 'SPRING_FLYWAY_ENABLED', 'true'
   environment 'SPRING_PROFILES_ACTIVE', 'dev'
+  // No LaunchDarkly key for local/cftlibTest - run the client offline so it serves code defaults
+  environment 'LAUNCHDARKLY_OFFLINE', 'true'
```
scottstewart-sl, line 555: Minor - you might want to set this for the generateCcdConfig task as well, as I have a feeling that this setting won't apply to it. It looks like the LD client went into offline mode anyway for that task as there is no SDK key, but you might want to be more explicit about it

```
2026-06-29T07:15:51.023 WARN  [main] u.g.h.reform.pcs.config.LaunchDarklyConfiguration
Starting LaunchDarkly client in offline mode (offlineMode=false, sdkKeyPresent=false) - serving code defaults
```

src/main/resources/application.yaml
```diff
@@ -237,6 +237,12 @@ access-code:
     max-retries: ${ACCESS_CODE_MAX_RETRIES:5}
     backoff-delay-seconds: ${ACCESS_CODE_BACKOFF_DELAY_SECONDS:3600s}
 
+launchdarkly:
+  sdk-key: ${LAUNCHDARKLY_SDK_KEY:}
+  offline-mode: ${LAUNCHDARKLY_OFFLINE:false}
+  env: ${LAUNCHDARKLY_ENV:default}
+  file: ${LAUNCHDARKLY_FILE:}
```
scottstewart-sl: Is this a list of 0 or more files? If so, could you call the property `files` rather than the singular `file`?

src/main/java/uk/gov/hmcts/reform/pcs/config/LaunchDarklyConfiguration.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private Optional<Path> getPathIfExists(String file) {
+        if (StringUtils.isBlank(file)) {
+            return Optional.empty();
+        }
+        Path flagFile = file.startsWith("/") ? Paths.get(file) : Paths.get("").resolve(file);
+        if (Files.exists(flagFile)) {
+            return Optional.of(flagFile);
+        }
+        log.debug("Could not find LaunchDarkly flag file defined by {}", file);
```
scottstewart-sl: Minor - probably this should be a warn, if it means that a dev has specified a local LD feature flag file but it couldn't be found

src/main/resources/application.yaml
arun-hmcts: Done

scottstewart-sl (approved):

PR #1778, opened 2026-05-05
Fallback to default uncategorised folder and remove Update access for documents

adusumillipraveen:
### Jira link

https://tools.hmcts.net/jira/browse/HDPI-6336

### Change description

Only realised in AAT that old cases where category id is not set, XUI by default adds a "Uncategorised documents" folder.  We think its better to fall back to that default behaviour to avoid having 2 uncategorised folders in the UI.

### Checklist

- [ ] commit messages are meaningful and follow good commit message guidelines
- [ ] README and other documentation has been updated / added (if needed)
- [ ] tests have been updated / new tests has been added (if needed)
- [ ] Does this PR introduce a breaking change

PR #2100, opened 2026-06-29
HDPI-7308: select document screen

sadmanrahh:
### Jira link

See [HDPI-7308](https://tools.hmcts.net/jira/browse/HDPI-7308)

### Change description

Add Select a document screen

image

### Testing done

  * [ ] Yes
  * [X] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/CaseFileCategory.java
```diff
@@ -11,10 +11,11 @@ public enum CaseFileCategory {
     PROPERTY_DOCUMENTS("propertyDocuments", "Property documents", 2),
     EVIDENCE("evidence", "Evidence", 3),
     HEARING_DOCUMENTS("hearingDocuments", "Hearing documents", 4),
-    ORDERS_AND_NOTICE_OF_HEARINGS("ordersAndNoticeOfHearings", "Orders and notice of hearings", 5),
+    ORDERS_AND_NOTICE_OF_HEARINGS("ordersAndNoticeOfHearings", "Orders and Notice of Hearings", 5),
     APPLICATIONS("applications", "Applications", 6),
     APPEALS("appeals", "Appeals", 7),
-    CORRESPONDENCE("correspondence", "Correspondence", 8);
+    CORRESPONDENCE("correspondence", "Correspondence", 8),
+    UNCATEGORISED_DOCUMENTS("uncategorisedDocuments", "Uncategorised documents", 9);
```
scottstewart-sl, line 19: This might end up with two Uncategorised documents folders in Case File View, if there is a document in the DB with no category. Worth testing in your PR environment. See https://github.com/hmcts/pcs-api/pull/1778 for some background.

Maybe an alternative is to make the default for the category column be `UNCATEGORISED_DOCUMENTS` so that there are never any null values in that column in the DB

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/documentamend/DocumentAmendDetails.java
```diff
@@ -0,0 +1,160 @@
...
+public class DocumentAmendDetails {
+
+    @CCD(
+        label = "Which folder is the document in?",
+        searchable = false,
+        typeOverride = FixedList,
+        typeParameterOverride = "DocumentAmendFolder"
+    )
+    @JsonProperty("SelectedFolder")
+    private DocumentAmendFolder selectedFolder;
```
scottstewart-sl: Is there a reason why you can't use `CaseFileCategory` here?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/documentamend/SelectDocumentPage.java
```diff
@@ -0,0 +1,199 @@
...
+        FieldCollectionBuilder<DocumentAmendDetails, State, ?> documentAmendPage = pageBuilder
+            .page(PAGE_ID, this::midEvent)
+            .pageLabel("Select document")
+            .label(PAGE_ID + "-separator", "---")
+            .complex(PCSCase::getDocumentAmendDetails)
+                .readonly(DocumentAmendDetails::getPropertyAddressSummary, NEVER_SHOW, true)
+                .mandatory(DocumentAmendDetails::getSelectedFolder)
+                .label("emptyFolderDocumentError", "", NEVER_SHOW)
+                .label("selectedFolderEmptyErrorMessage", "", NEVER_SHOW)
+                .label("emptyFolderDocumentQuestion", "", NEVER_SHOW);
```
scottstewart-sl, line 42: Minor - you could stick a `.done()` on the end here to close off the `complex()` block

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/documentamend/SelectDocumentPage.java
```diff
@@ -0,0 +1,199 @@
...
+
+    private String documentQuestion() {
+        return "Which document do you want to amend?";
+    }
+
+    private String errorMessage(String message) {
+        return "<p class=\"govuk-error-message\">" + message + "</p>";
+    }
+
+    private enum DocumentCategoryField {
```
scottstewart-sl: I think it would be slightly cleaner to pop this enum out into its own file.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/documentamend/SelectDocumentPage.java
sadmanrahh: I’ve left this as-is because the following category fields are also part of documentAmendDetails, so the complex() block needs to stay open until those fields and hidden readonly values are registered. The .done() at the end closes the block after all DocumentAmendDetails fields have been added.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/documentamend/DocumentAmendDetails.java
sadmanrahh: Good point, there wasn't enough value in keeping a separate DocumentAmendFolder wrapper. I've removed it and now use CaseFileCategory directly for the selected folder, with CaseFileCategory implementing HashLabel so the CCD dropdown labels remain unchanged.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/documentamend/SelectDocumentPage.java
sadmanrahh: Sure, addressed

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/documentamend/DocumentAmendDetails.java
```diff
@@ -0,0 +1,160 @@
...
+
+    @CCD(searchable = false)
+    @JsonProperty("PropertyAddressSummary")
```
scottstewart-sl: Minor - you can use `@JsonNaming(PropertyNamingStrategies.UpperCamelCaseStrategy.class)` on the class to prevent having to add these JsonProperty annotations to each field. (And that's more consistent with what we do in some of the other domain classes)

ayub-macalim (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/documentamend/DocumentAmendDetails.java
sadmanrahh: Good point, updated

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/CaseFileCategory.java
sadmanrahh: Good catch, I tested this and confirmed null category_id plus explicit uncategorisedDocuments creates duplicate uncategorised folders in Case File View. I’ve added a migration to backfill null categories and set a DB default, and updated document persistence so new documents default to uncategorisedDocuments instead of saving null.

src/main/resources/db/migration/V135__default_uncategorised_document_category.sql
```diff
@@ -0,0 +1,6 @@
+UPDATE document
+SET category_id = 'uncategorisedDocuments'
+WHERE category_id IS NULL;
+
+ALTER TABLE document
+ALTER COLUMN category_id SET DEFAULT 'uncategorisedDocuments';
```
scottstewart-sl, line 6: Minor - does this need a NOT NULL constraint as well?

scottstewart-sl (approved):

src/main/resources/db/migration/V135__default_uncategorised_document_category.sql
sadmanrahh: I checked the insert paths and there is still a DEFENDANT_ACCESS_CODE in DefendantAccessCodeService that currently saves category_id as null. If we add NOT NULL, we’ll need to decide whether those documents should be categorised, for example as uncategorisedDocuments, or excluded from Case File View. Do you think we should handle that in this PR and then add NOT NULL, or keep this change limited to the backfill/default for now?

PR #2113, opened 2026-07-01
Hdpi 5947 bulk print

arun-hmcts:
### Jira link

[HDPI-5947](https://tools.hmcts.net/jira/browse/HDPI-5947)

### Change description

- Adds a scheduled, LaunchDarkly-gated sweep that coversheets generated documents and posts them via Send Letter — claim pack (claimant + defendants, each with their own access code) and defence pack (defence form + counter-claim, fanned to all parties).
- Per-(party, document) send tracking so re-runs never double-post and failures self-heal on the next sweep; missing-address letters are skipped and retried.
- PdfMerger merges coversheet + documents into one PDF so Send Letter can't drop a genuinely different pack sharing the identical coversheet, while exact retries still de-duplicate.
- Per-environment schedule (prod nightly, AAT hourly, PR 60s) with a lookback cutoff to bound the backlog; off by default to avoid postage cost.
https://tools.hmcts.net/confluence/spaces/RRFM/pages/1990363540/How+it+will+work

### Testing done

Tested in PR ..except counter claim form. Once peer review completed , will be testing it after merging the changes into this branch.

  * [ ] Yes
  * [x] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/DefencePackSender.java
```diff
@@ -0,0 +1,79 @@
...
+        this.recipientAddressResolver = recipientAddressResolver;
+        this.addressMapper = addressMapper;
+        this.bulkPrintService = bulkPrintService;
+        this.packSendRecorder = packSendRecorder;
+    }
+
+    @Transactional
+    public void sendDefencePacks(UUID caseId) {
+        pcsCaseRepository.findById(caseId).ifPresent(pcsCase ->
+            defencePackSelector.findDefencePackCandidates(pcsCase)
```
tvr-hmcts: I've a feeling that this loop of remote calls under Transaction could be a problem.

tvr-hmcts:
I think I am seeing 6 remote calls in loops under db transactions here.

AccessCodeGenerationService.java
ClaimPackSender.java
BulkPrintService.java
BulkPrintScheduledTask.java (twice)
DefencePackSender.java

src/main/resources/db/migration/V133__add_defence_pack_activity_types.sql
```diff
@@ -0,0 +1,4 @@
+-- Per-document send tracking: one DOCUMENT_SENT row per (recipient, document).
+ALTER TYPE claim_activity_type ADD VALUE IF NOT EXISTS 'DOCUMENT_SENT';
+ALTER TABLE claim_activity_log ADD COLUMN document_id UUID REFERENCES document(id);
+CREATE INDEX IF NOT EXISTS idx_claim_activity_log_type_status ON claim_activity_log (activity_type, status);
```
scottstewart-sl: What do you think about a clean up of the existing data here, so we can remove the old enum values in a subsequent PR?

src/main/resources/application.yaml
```diff
@@ -263,6 +263,24 @@ access-code:
     max-retries: ${ACCESS_CODE_MAX_RETRIES:5}
     backoff-delay-seconds: ${ACCESS_CODE_BACKOFF_DELAY_SECONDS:3600s}
 
+bulk-print:
+  schedule: ${BULK_PRINT_SCHEDULE:DAILY|02:00}
+  # BULK_PRINT_LOOKBACK_HOURS (lower envs only): sweep ignores packs whose documents are older than N hours,
+  # so enabling the flag cannot flood send-letter with the historical backlog. Unset in prod = full backlog.
+
+send-letter:
+  url: ${SEND_LETTER_URL:http://rpe-send-letter-service-aat.service.core-compute-aat.internal}
```
scottstewart-sl: Minor - it's better to make the default something on localhost, (and set in build.gradle if necessary for local dev). That way it would fail in AAT if `SEND_LETTER_URL` wasn't set, rather than only discovering the issue in Demo/Perftest

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/BulkPrintScheduledTask.java
```diff
@@ -0,0 +1,89 @@
...
+            MDC.remove(MDC_TASK_NAME);
+        }
+    }
+
+    private List<UUID> discoverCandidateCases() {
+        if (lookbackHours == null) {
+            return claimActivityLogRepository.findCaseIdsByActivityTypeAndStatus(
+                ClaimActivityType.DOCUMENTS_CREATED, ClaimActivityStatus.SUCCESS);
+        }
+        LocalDateTime cutoff = LocalDateTime.now().minusHours(lookbackHours);
```
scottstewart-sl: Have you double checked this cutoff when the JVM is running in UK time, (BST) and the DB is in UTC?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/form/DefenceCorrespondenceAddressResolver.java
```diff
@@ -0,0 +1,66 @@
...
+        AddressEntity fallback = defendant.getAddress() != null ? defendant.getAddress() : propertyAddress;
+        return fallback == null ? null : addressMapper.toAddressUK(fallback);
+    }
+
+    private AddressUK assertedCorrespondenceAddress(PartyEntity defendant) {
+        String assertedValue = assertionRepository.findByPartyIdAndAssertedByAndStatus(
+                defendant.getId(),
+                PartyAttributeAssertedBy.DEFENDANT,
+                PartyAttributeAssertionStatus.SUBMITTED).stream()
+            .filter(assertion -> assertion.getAttributesName() == PartyAttributeType.CORRESPONDENCE_ADDRESS)
```
scottstewart-sl, line 51: Just double checking - is this correct that the defendant's asserted address is used for correspondence before it has PartyAttributeAssertionStatus.ACCEPTED? I can't see anything about that specified in the ticket, so perhaps check with the BA if it isn't specified elsewhere.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/BulkPrintScheduledTask.java
```diff
@@ -0,0 +1,89 @@
...
+            log.info("Bulk print sweep starting for {} candidate cases (lookbackHours={})",
+                caseIds.size(), lookbackHours);
+            caseIds.forEach(caseId -> {
+                claimPackSender.sendClaimPacks(caseId);
+                defencePackSender.sendDefencePacks(caseId);
+            });
+        } finally {
+            MDC.remove(MDC_TASK_NAME);
+        }
+    }
```
gergelykisshmcts, line 76: Just checking the scheduler discovery logic: are we confident every eligible case will always get a DOCUMENTS_CREATED/SUCCESS activity log when the docs are ready?
Also confirming prod has no lookback limit, so failed/held sends cannot age out and be missed later.
My only concern is docs existing but the case no longer being picked up by the scheduler, which would be hard to diagnose.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/BulkPrintService.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    public UUID sendPack(PcsCaseEntity pcsCase, PartyEntity recipient, LetterType letterType,
+                         String recipientName, AddressUK address, List<DocumentEntity> documents) {
+        requirePostalAddress(address, recipient);
+        String caseReference = caseReferenceFormatter.formatCaseReferenceWithDashes(pcsCase.getCaseReference());
+
+        List<byte[]> pdfs = new ArrayList<>();
+        pdfs.add(coversheetProvider.render(recipientName, address, caseReference));
+        documents.forEach(document -> pdfs.add(letterDocumentFetcher.fetchBytes(document.getDocumentId())));
```
tvr-hmcts, line 62: This can potentially be multi-threaded.

Wondering then also what scope there is to stream these into temp local files and then merging from that may be more efficient.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/LetterDocumentFetcher.java
```diff
@@ -0,0 +1,51 @@
...
+            systemUpdateUserTokenProvider.getAuthToken(),
+            authTokenGenerator.generate(),
+            documentId
+        ).getBody();
+        return readAllBytes(resource);
+    }
+
+    private byte[] readAllBytes(Resource resource) {
+        try {
+            return resource.getInputStream().readAllBytes();
```
tvr-hmcts: Think you need to close the stream

src/main/resources/application.yaml
arun-hmcts: Fixed

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/BulkPrintScheduledTask.java
arun-hmcts: Fixed

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/BulkPrintService.java
```diff
@@ -0,0 +1,98 @@
...
+        Document mergedLetter = new Document(Base64.getEncoder().encodeToString(pdfMerger.merge(pdfs)), SINGLE_COPY);
```
tvr-hmcts, line 63: If the local temp file approach is take then this pdfMerger.merge(...) could just pass in the reference to the files and use loadXFDF(...)

That will make the api calls more performant as they will be both threaded and also streaming to files so outside of hte heap space.

Then pdfMerger.merge() can use the files to merge.

Tidying up the files at the end or having a background task do it.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/BulkPrintScheduledTask.java
```diff
@@ -0,0 +1,89 @@
...
+            return;
+        }
+        MDC.put(MDC_TASK_NAME, BULK_PRINT_TASK_NAME);
+        try {
+            List<UUID> caseIds = discoverCandidateCases();
+            log.info("Bulk print sweep starting for {} candidate cases (lookbackHours={})",
+                caseIds.size(), lookbackHours);
+            caseIds.forEach(caseId -> {
+                claimPackSender.sendClaimPacks(caseId);
+                defencePackSender.sendDefencePacks(caseId);
```
tvr-hmcts: Is there a dependency between these two?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/PackSendRecorder.java
```diff
@@ -0,0 +1,96 @@
...
+    private static final String MDC_DOCUMENT_ID = "documentId";
+    private static final String MDC_DOCUMENT_TYPE = "documentType";
+
+    private final AccessCodeActivityLogService accessCodeActivityLogService;
+
+    public PackSendRecorder(AccessCodeActivityLogService accessCodeActivityLogService) {
+        this.accessCodeActivityLogService = accessCodeActivityLogService;
+    }
+
+    public void send(PcsCaseEntity pcsCase, PartyEntity recipient, LetterType letterType,
```
tvr-hmcts: Not sure on the name of this method as it does not send but instead records that something has been sent.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/BulkPrintService.java
```diff
@@ -0,0 +1,98 @@
...
+
+        LetterV3 letter = new LetterV3(letterType.getCode(), List.of(mergedLetter),
+            additionalData(caseReference, recipientName, address));
+        return sendLetterApi.sendLetter(authTokenGenerator.generate(), letter).letterId;
```
tvr-hmcts, line 67: Once threaded up and an merged pdf has been generated which can be held temp also then this can be triggered to run and send it.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/BulkPrintService.java
```diff
@@ -0,0 +1,98 @@
...
+        this.caseReferenceFormatter = caseReferenceFormatter;
+    }
+
+    public UUID sendPack(PcsCaseEntity pcsCase, PartyEntity recipient, LetterType letterType,
+                         String recipientName, AddressUK address, List<DocumentEntity> documents) {
+        requirePostalAddress(address, recipient);
+        String caseReference = caseReferenceFormatter.formatCaseReferenceWithDashes(pcsCase.getCaseReference());
+
+        List<byte[]> pdfs = new ArrayList<>();
+        pdfs.add(coversheetProvider.render(recipientName, address, caseReference));
```
tvr-hmcts, line 61: Noticed this remote call after the other comments so this can be added to the threading and written out to a temp file.

tvr-hmcts (commented):
Some thoughts/comments

gergelykisshmcts (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/LetterDocumentFetcher.java
tvr-hmcts: Is the size of the stream known?  If cut then an exception would not happen and it is possible only half the file can get returned.

Perhaps a check on 0 bytes also.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/BulkPrintScheduledTask.java
```diff
@@ -0,0 +1,100 @@
...
+            MDC.remove(MDC_TASK_NAME);
+        }
+    }
+
+    // Isolate each case: a failure discovering or sending one case's packs is logged and skipped, so one bad
+    // case never aborts the sweep for the rest. Per-recipient failures are already handled inside the senders.
+    private void sendPacksForCase(UUID caseId) {
+        try {
+            claimPackSender.sendClaimPacks(caseId);
+            defencePackSender.sendDefencePacks(caseId);
```
tvr-hmcts: Is it still relevant to call the defencePackSender in the scenario when the pdf merge has not completed successfully?

tvr-hmcts (commented):
Couple of other thoughts.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/form/DefenceCorrespondenceAddressResolver.java
arun-hmcts: Its an requirement from the service team .. that address provided by defendant name and address should be used

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/BulkPrintScheduledTask.java
arun-hmcts: Yes, confident on all three. The claim form document and its DOCUMENTS_CREATED/SUCCESS row are written in the same transaction (ClaimFormPersistenceService.attach / DefendantAccessCodeService.generateForDefendant, both @Transactional), so a committed document can never exist without its discovery row — I've added an integration test (ClaimFormPersistenceServiceIT) that proves it: forcing the activity-log write to fail rolls back the document. Prod has no lookback, so once a case has that row it's re-checked on every sweep and, with per-(recipient, document) tracking, any held/failed send is retried indefinitely — nothing ages out. So the "docs exist but the case isn't picked up" scenario can't happen through the code, and it's now regression-guarded by the test.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/LetterDocumentFetcher.java
arun-hmcts: Fixed — readAllBytes now uses try-with-resources (try (InputStream in = resource.getInputStream())) so the stream is always closed.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/BulkPrintScheduledTask.java
arun-hmcts: Fixed — claim and defence are now sent in independent, isolated phases (sendPhase), so a failure in one no longer skips the other or the rest of the sweep.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/PackSendRecorder.java
arun-hmcts: Renamed to sendAndRecord — it runs the send action and records the per-document outcome, so the name now reflects both.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/BulkPrintService.java
arun-hmcts: R1A packs are small — coversheet + one or two small documents — so heap usage per letter is modest. Let us take its perf optimisation later on.Thank you

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/BulkPrintService.java
arun-hmcts: since at R1A pack sizes the in-memory path is fine for now.. later we can take it up later.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/LetterDocumentFetcher.java
arun-hmcts: done

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/BulkPrintScheduledTask.java
arun-hmcts: Yes — claim and defence run as independent isolated phases now, and a PDF merge failure is per-recipient (recorded and retried next sweep), so a claim-side failure never stops defence.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/BulkPrintService.java
arun-hmcts: For later for optimisation ..

scottstewart-sl (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/LetterDocumentFetcher.java
```diff
@@ -0,0 +1,63 @@
...
+        this.systemUpdateUserTokenProvider = systemUpdateUserTokenProvider;
+    }
+
+    public byte[] fetchBytes(UUID documentId) {
+        ResponseEntity<Resource> response = caseDocumentClientApi.getDocumentBinary(
+            systemUpdateUserTokenProvider.getAuthToken(),
+            authTokenGenerator.generate(),
+            documentId
+        );
+        byte[] content = readAllBytes(response.getBody());
```
tvr-hmcts, line 43: The response could be null (less likely) but the response body could however be null.   Also should the response HTTP status be checked first?  Guessing this should be a 20X response with a body.

src/main/java/uk/gov/hmcts/reform/pcs/config/SchedulingConfig.java
```diff
@@ -59,12 +60,24 @@ public Scheduler startupTasksScheduler(DataSource dataSource,
...
+            .filter(task -> !(task instanceof RecurringTask))
+            .toList();
+
+        var builder = Scheduler.create(dataSource, oneTimeTasks)
             .threads(threadCount)
             .pollingInterval(Duration.ofSeconds(interval))
             .serializer(new JacksonSerializer(objectMapper))
             .registerShutdownHook();
 
+        for (Task<?> task : tasks) {
```
tvr-hmcts, line 75: Minor but wondering if there should be some coverage added for this class.  I recognise it has not been there before though.

src/main/java/uk/gov/hmcts/reform/pcs/config/LaunchDarklyConfiguration.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private ComponentConfigurer<DataSource> getDataSource(Path[] flagFilePaths) {
+        return FileData.dataSource()
+            .filePaths(flagFilePaths)
+            .duplicateKeysHandling(FileData.DuplicateKeysHandling.IGNORE);
+    }
+
+    private Optional<Path[]> getExistingFiles(String[] files) {
+        if (files == null || files.length < 1) {
+            return Optional.empty();
```
tvr-hmcts, line 55: None of the tests bump into this.

src/main/java/uk/gov/hmcts/reform/pcs/config/LaunchDarklyConfiguration.java
```diff
@@ -0,0 +1,75 @@
...
+            .map(this::getPathIfExists)
+            .filter(Optional::isPresent)
+            .map(Optional::get)
+            .toArray(Path[]::new);
+        return existing.length > 0 ? Optional.of(existing) : Optional.empty();
+    }
+
+    private Optional<Path> getPathIfExists(String file) {
+        if (StringUtils.isBlank(file)) {
+            return Optional.empty();
```
tvr-hmcts, line 67: No tests hit this.

src/main/java/uk/gov/hmcts/reform/pcs/config/LaunchDarklyConfiguration.java
```diff
@@ -0,0 +1,75 @@
...
+        }
+        Path flagFile = file.startsWith("/") ? Paths.get(file) : Paths.get("").resolve(file);
+        if (Files.exists(flagFile)) {
+            return Optional.of(flagFile);
+        }
+        log.warn("Could not find LaunchDarkly flag file defined by {}", file);
+        return Optional.empty();
```
tvr-hmcts, line 74: or this so not sure the method gets hit by the tests.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/AccessCodeActivityLogService.java
```diff
@@ -54,4 +94,17 @@ private void save(PcsCaseEntity pcsCase,
                   pcsCase.getId(),
                   party != null ? " and party " + party.getId() : "");
     }
+
+    private String toJson(ActivityDetails details) {
+        if (details == null) {
+            return null;
+        }
+        try {
+            return objectMapper.writeValueAsString(details);
```
tvr-hmcts, line 103: All the tests return null so this is not covered or the exception.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/form/DefenceCorrespondenceAddressResolver.java
```diff
@@ -0,0 +1,66 @@
...
+            .findFirst()
+            .orElse(null);
+        if (assertedValue == null) {
+            return null;
+        }
+        try {
+            return objectMapper.readValue(assertedValue, AddressUK.class);
+        } catch (Exception e) {
+            log.error("Failed to parse defendant correspondence address assertion", e);
+            return null;
```
tvr-hmcts, line 63: Minor as the behaviour will be the same from the null return above but the tests do not reach here.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/defenceform/DefenceFormPayloadBuilder.java
```diff
@@ -382,20 +386,11 @@ private String resolveDefendantName(PartyEntity defendant,
...
-                    objectMapper.readValue(assertedAddress, AddressUK.class), NEWLINE_DELIMITER);
-            } catch (Exception e) {
-                log.error("Failed to parse defendant correspondence address assertion", e);
-            }
-        }
-        return formatAddress(defendant.getAddress() != null ? defendant.getAddress() : propertyAddress);
+    private String resolveDefendantAddress(PartyEntity defendant, AddressEntity propertyAddress) {
+        AddressUK address =
+            defenceCorrespondenceAddressResolver.resolveCorrespondenceAddress(defendant, propertyAddress);
+        return address == null ? null
```
tvr-hmcts, line 392: Address is always not null within the tests.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/claimform/ClaimActivityLogService.java
```diff
@@ -27,45 +33,62 @@
...
+    private String toJson(GenerationDetails details) {
+        if (details == null) {
+            return null;
+        }
+        try {
+            return objectMapper.writeValueAsString(details);
+        } catch (JsonProcessingException e) {
+            // Never fail the activity write over its detail payload; the row itself is the important record.
+            log.error("Failed to serialise generation details {}", details, e);
+            return null;
```
tvr-hmcts, line 88: Minor as the same behaviour not doubt from the GenerationDetails being null itself but this path is not covered in the tests.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/SentPackDocuments.java
```diff
@@ -0,0 +1,58 @@
...
+ * Derives which (recipient, document) pairs have already been posted from the {@code PACK_SENT}
+ * SUCCESS rows' {@link PackDetails#documents()} — the bulk-print dedup source. A pack's documents
+ * are merged into one letter and posted atomically, so pack coverage and per-document coverage are
+ * equivalent; a document not covered by any prior pack (e.g. a late counter-claim) goes out on the
+ * next sweep without re-sending the rest.
+ */
+@Service
+@RequiredArgsConstructor
+@Slf4j
+public class SentPackDocuments {
```
tvr-hmcts, line 27: Probably should have coverage on this class.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/PackSendRecorder.java
```diff
@@ -0,0 +1,106 @@
...
+    private static final String MDC_TERMINAL_FAILURE = "terminalFailure";
+    private static final String MDC_FAILURE_REASON = "failureReason";
+
+    private final AccessCodeActivityLogService accessCodeActivityLogService;
+
+    public PackSendRecorder(AccessCodeActivityLogService accessCodeActivityLogService) {
+        this.accessCodeActivityLogService = accessCodeActivityLogService;
+    }
+
+    public void sendAndRecord(PcsCaseEntity pcsCase, PartyEntity recipient, LetterType letterType,
```
tvr-hmcts, line 45: Probably needs coverage.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/PackSendRecorder.java
```diff
@@ -0,0 +1,106 @@
...
+            accessCodeActivityLogService.recordPackSent(pcsCase, recipient,
+                PackDetails.sent(letterType, packDocumentRefs(documents), letterId));
+            log.info("Pack sent - case: {}, party: {}, letterType: {}, letterId: {}, documents: {}",
+                pcsCase.getCaseReference(), recipient.getId(), letterType.getCode(), letterId,
+                documentSummary(documents));
+            documents.forEach(document -> logDocumentSent(pcsCase, recipient, document, letterType, letterId));
+        } catch (MissingPostalAddressException e) {
+            recordFailure(pcsCase, recipient, letterType, documents, e, true);
+        } catch (Exception e) {
+            recordFailure(pcsCase, recipient, letterType, documents, e, false);
```
tvr-hmcts, line 62: No tests hit this from what I can tell

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/bulkprint/PackRecipientResolver.java
```diff
@@ -0,0 +1,95 @@
...
+    private AddressUK correspondenceAddress(PartyEntity recipient, PartyRole role, AddressEntity propertyAddress) {
+        if (role == PartyRole.DEFENDANT) {
+            return defenceCorrespondenceAddressResolver.resolveCorrespondenceAddress(recipient, propertyAddress);
+        }
+        return postalAddress(recipient, role, propertyAddress);
+    }
+
+    private AddressUK postalAddress(PartyEntity recipient, PartyRole role, AddressEntity propertyAddress) {
+        AddressEntity resolved = recipientAddressResolver.resolvePostalAddress(recipient, role, propertyAddress);
+        return resolved == null ? null : addressMapper.toAddressUK(resolved);
```
tvr-hmcts, line 89: Tests alway have a resolved address.

tvr-hmcts (commented):
Some minors but one potential nullpointers with the need to check on the HTTP status also

PR #2123, opened 2026-07-02
HDPI-5819: Restore testing-support enabled default to true

arun-hmcts:
### Jira link

House Keeping .
See [PROJ-XXXXXX](https://tools.hmcts.net/jira/browse/PROJ-XXXXXX)

### Change description

 HDPI-5819 changed the testing-support default to false and that went out to AAT today, which switched off all the /testing-support/** endpoints there. AAT never sets ENABLE_TESTING_SUPPORT in its flux config, so it was relying on the default being true — so the moment the default flipped, testing-support disappeared on AAT and the frontend smoke tests (and our own FeePaymentCallbackTests) started failing because they set up their cases and payments through those endpoints.

This just puts the default back to true so AAT works again while we sort the proper fix.

  * [ ] Yes
  * [ ] No

### Checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -543,6 +543,7 @@ tasks.withType(CftlibExec).configureEach {
     logger.quiet("Attaching Mockito Java agent for inline mocking")
   }
 
+  environment 'ENABLE_TESTING_SUPPORT', 'true'
   environment 'RSE_LIB_ADDITIONAL_DATABASES', 'pcs'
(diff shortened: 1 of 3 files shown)


scottstewart-sl (approved):

benouaer (approved):

sdmiddletonSolirius (approved):
LGTM

fudge88 (approved):

PR #2155, opened 2026-07-09
HDPI-7579: Control caseworker "change case state" event with feature flags

scottstewart-sl:
### Jira link

See [HDPI-7579](https://tools.hmcts.net/jira/browse/HDPI-7579)

### Change description

Control caseworker "change case state" event with feature flags. Also adds a re-usable show condition helper that builds a feature flag based show condition from one or more feature flags.

### Testing done

Manual testing locally and in preview, changing the feature flags and confirming that the event was visible or hidden in the Next Steps dropdown as expected

  * [ ] Yes
  * [x] No

### Checklist


libanAbdirahman1 (approved):

src/test/java/uk/gov/hmcts/reform/pcs/ccd/ShowConditionsTest.java
```diff
@@ -56,10 +64,48 @@ void shouldJoinShowConditionWithAnd(List<String> showConditionsToJoin, String ex
...
+        // When
+        String actualShowCondition = ShowConditions.featureFlagsEnabled(featureFlags.toArray(new FeatureFlag[0]));
+
+        // Then
+        assertThat(actualShowCondition).isEqualTo(expectedShowCondition);
+    }
+
+    @ParameterizedTest
+    @EnumSource(value = FeatureFlag.class, names = {"RELEASE_1_DOT_2", "CASEWORKER_EVENTS"}, mode = EXCLUDE)
+    void shouldThrowExceptionForFeatureFlagWithNoCcdField(FeatureFlag featureFlag) {
```
tvr-hmcts, line 97: Am I understanding the 'hidden' intention here correctly.   This test will fail when new flags are added but not fully applied?

Therefore enforcing the developer to complete the task to get the test to pass.

If so.  That's awesome :)

scottstewart-sl: Not quite that clever... Not all feature flags will be used to control features in ExUI, (for example bulk print isn't), so this test is checking the functionality that throws an exception if a developer tries to configure a show condition using one of the `FeatureFlag` values that isn't mapped to a CCD field (yet).

For completeness, I've just added another test which is the opposite of this, and asserts that an exception is not thrown for the `FeatureFlag` values that should have a CCD field mapping.

tvr-hmcts: Ah ok thank you for the explanation :)

Does that mean then that there should be a FlagType of some sort to distinguish them then within the FeatureFlag enum constructor.

Then it is clear to someone reading the FeatureFlag enum class as to what is relevant where and also then these tests could perhaps use that to distinguish?

scottstewart-sl: I don't think that is necessary at the moment. If a dev wants to use a FeatureFlag in a show condition and it's not already mapped to a CCD field then an exception will be thrown when the CCD config is generated, and they can resolve the issue at that stage.

tvr-hmcts: Ok cool.   Was just thoughts while in the area.

tvr-hmcts (approved):

PR #1978, opened 2026-06-12
HDPI-7052: add GroupAccess CCD definition config

benouaer:
### Jira link
- See [HDPI-7052](https://tools.hmcts.net/jira/browse/HDPI-7052)

### Change description
- PR created to test GroupAccess configuration  in preview

### Testing done
--

### Security Vulnerability Assessment ###
**CVE Suppression:** Are there any CVEs present in the codebase (either newly introduced or pre-existing) that are being intentionally suppressed or ignored by this commit?
  * [ ] Yes
  * [x] No

### Checklist
- [x] commit messages are meaningful and follow good commit message guidelines
- [ ] README and other documentation has been updated / added (if needed)
- [ ] tests have been updated / new tests has been added (if needed)
- [ ] Does this PR introduce a breaking change

PR #2179, opened 2026-07-15
HDPI-7052: organisation-wide case access, visible from draft creation

arun-hmcts:
### Jira link

See [HDPI-7052](https://tools.hmcts.net/jira/browse/HDPI-7052)

### Change description

An organisation can now see and work on its own cases, including drafts, from the moment one is created.

Two things have to match: PRM gives each user a role carrying a `caseAccessGroupId`, and the case has to carry the same id in `CaseAccessGroups`. The spike branch (#1978) declares the access types that produce the roles; this PR derives the group onto the case, and does it at creation so drafts are covered rather than only submitted cases.

#### Deriving the group onto the case

- `CaseAccessGroupsUtil` builds the group id from the same `GroupAccessType` enum the definition comes from, so the two sides cannot drift. Applied on read by `PCSCaseView`; `party.organisation_profile_ids` added (V012)
- `PartyService.initialiseClaimant` gives a new case its claimant organisation immediately, so the group derives on a draft. Submit completes that party rather than adding a second
- An organisation is now required to create a case. Without one nothing derives, and after `[CREATOR]` is revoked at submit the case would be unreachable
- `[CREATOR]` is granted by CCD inside the create transaction, so the create-time role task is removed; PCS only revokes, at submit

#### Drafts now belong to the firm

A claim draft was keyed on `idam_user_id`. A colleague at the same firm opened the case and got a blank form, with nothing to show why. Claim drafts are now owned by the case's organisation-owned claimant party instead.

- `owner_party_id` added to `draft.draft_case_data` (V012), with a partial unique index on `(case_reference, event_id, owner_party_id)` where it is not null. One draft per firm per event is now enforced by the database, and per-user rows are left alone
- `DraftOwnership` classifies each event: `PARTY` for `resumePossessionClaim`, `USER` for `respondPossessionClaim` and `enforceTheOrder`, `NO_DRAFTS` for the rest. The user-keyed lookup also requires `owner_party_id IS NULL`, so it can't pick up a shared draft
- `respondPossessionClaim` stays per-user on purpose. HDPI-6221 exists so a citizen and their legal rep don't share a draft
- `idam_user_id` is still written, so a shared draft records who last touched it

The failure mode here is silent, so there are three guards. A static initialiser won't let the class load if any `EventId` is unclassified, because an unclassified event would quietly fall back to a per-user draft. Resolving the owner throws if there are no organisation-owned claimant parties, or more than one, rather than falling back to the user key. And a `NO_DRAFTS` event that reaches the draft service throws instead of inventing a draft.

#### Access control

- `ClaimantAccess` no longer grants `caseworker-pcs-solicitor`. That blanket IDAM role was standing in for the group roles, which meant claimant data showed up whether or not group access resolved, and the two-user test could not fail
- `createPossessionClaim` granted to the `claimant` capacity: local authorities and the "other" profiles are the claimant themselves and could resume but not create. `caseworker-pcs-solicitor` is kept there for create only, since a case has no `CaseAccessGroups` until it has a claimant party, so the group roles can't reach the event that creates one
- `[SOLICITOR]` removed. Nothing ever assigned it, and it was missing from `AccessProfile` so it never reached the CaseRoles tab. Its ~189 authorisation rows were failing the definition import
- Group role names use hyphens (`claimant-solicitor`, `defendant-solicitor`) to match the RAS catalogue. With underscores nothing matches and there is no error

SDK pinned to `access-group-roles-optin-126-126.1-51aae294` — [dtsse-ccd-config-generator#1063](https://github.com/hmcts/dtsse-ccd-config-generator/pull/1063) plus [`d088070c`](https://github.com/hmcts/dtsse-ccd-config-generator/commit/d088070c) (`CaseAccessGroup` type) and [`51aae294`](https://github.com/hmcts/dtsse-ccd-config-generator/commit/51aae294) (`grantCaseTypeAccessToCaseRoles`, without which `[CREATOR]` gets no case-type ACL).

#### Depends on

[am-role-assignment-service#2781](https://github.com/hmcts/am-role-assignment-service/pull/2781) (POFCC-368) defines the four PCS roles and hasn't merged yet. The preview pins `role-assignment-service:pr-2781` so it works here. AAT won't until that lands.

Known gaps: `defendant-solicitor` and `duty-advisor-request` are in `AccessTypeRole` but have no field authorisations, and derivation is claimant-only — the defendant side goes with notice of change. Two parties in one organisation produce duplicate groups. Access types are mandatory, so the Manage Organisation checkbox is advisory. `GroupAccessFields.caseAccessGroups` is `@CCD` with no `access`, so the stamp doesn't come back over the API. That doesn't affect access, but you have to read it from the database when something looks wrong.

### Testing done

Unit tests for the initialised claimant and the organisation precondition. `CaseAccessGroupsUtilTest` covers derivation including the pre-submission claimant, and `DraftCaseDataServiceTest` covers the party-owned lookup, the user-keyed fallback and the three guards. The create callback fixture asserts the case now carries a claimant, and the resume fixture still expects two parties, which catches a duplicate claimant.

Checked on the PR preview against the RAS `pr-2781` image, both branches of the config:

**Solicitor** (`solicitor-org-claimant-access` / `claimant-solicitor`). Case `1786602938415743` at Eviction Solicitor (`QKLHPMU`): creator 200, second user at the same firm 200 with no share and no case role, SUN SOLICITORS (`IHOVCKH`) 404, One Local Authority (`ZLQXW5K`) 404. All four users hold `caseworker-pcs-solicitor` and `caseworker-pcs`, and two of them are refused, so the blanket roles aren't what's granting access. The only difference is which organisation they're in.

**Local authority** (`prof-org-claimant-access` / `claimant`), which is the access type the other four profiles share. Case `1786603624292172` at One Local Authority (`ZLQXW5K`), draft saved by user1 and opened by user2. In the database the claimant party has `organisation_id=ZLQXW5K` and `{LOCALAUTH_PROFILE}`, and the draft row has `owner_party_id` pointing at that party with `party_id` null, so it is party-owned rather than user-keyed. The derived id and the one in the user's RAS assignment are identical (`PCS:PCS:prof-org-claimant-access:claimant:ZLQXW5K`), and user2 gets 200 and picks up user1's draft.

  * [x] Yes
  * [ ] No

CVE ID: CVE-2026-64607 (httpclient5 5.5.2, transitive via Spring/Feign). Dated suppression until 25/08/2026, pending an upstream release.

### Checklist

### PCS checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/GroupAccessType.java
```diff
@@ -102,6 +113,34 @@ public enum GroupAccessType implements CCDAccessGroup {
...
+    private static Map<Key, GroupAccessType> buildIndex() {
+        Map<Key, GroupAccessType> index = new HashMap<>();
+        for (GroupAccessType accessType : values()) {
+            if (accessType.partyRole == null) {
+                continue;
+            }
+            GroupAccessType clash = index.put(
+                new Key(accessType.organisationProfileId, accessType.partyRole), accessType);
+            if (clash != null) {
+                throw new IllegalStateException(
```
benouaer: do we need this? as this will never happen

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/GroupAccessFields.java
```diff
@@ -0,0 +1,24 @@
...
+import uk.gov.hmcts.ccd.sdk.type.ListValue;
+
+/**
+ * Notice of change owns the ChangeOrganisationRequest field, declared on {@link PCSCase}. The
+ * definition store allows only one per case type - a second declaration here failed the import with
+ * "Change Organisation Request is defined more than once for case type 'PCS'".
+ */
+@Data
+@Builder
+public class GroupAccessFields<R extends HasRole> {
```
benouaer: this class is not needed anymore, we can move caseAccessGroups field to PCSCase

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/GroupAccessType.java
```diff
@@ -102,6 +113,34 @@ public enum GroupAccessType implements CCDAccessGroup {
...
+    private record Key(String organisationProfileId, PartyRole partyRole) { }
+
+    private static Map<Key, GroupAccessType> buildIndex() {
+        Map<Key, GroupAccessType> index = new HashMap<>();
+        for (GroupAccessType accessType : values()) {
+            if (accessType.partyRole == null) {
+                continue;
+            }
+            GroupAccessType clash = index.put(
+                new Key(accessType.organisationProfileId, accessType.partyRole), accessType);
```
benouaer: we probably just need the caseAccessGroupIdTemplate, rather than the whole accessType here

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/PCSCase.java
```diff
@@ -788,6 +789,9 @@ public class PCSCase {
     @CCD(access = {DefendantSolicitorAccess.class})
     private List<ListValue<Party>> allLinkedDefendants;
 
+    @JsonUnwrapped
+    private GroupAccessFields<AccessProfile> groupAccessFields;
```
benouaer: we should just replace this caseAccessGroups

src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/party/PartyEntity.java
```diff
@@ -75,6 +75,7 @@ public class PartyEntity {
     private String orgName;
 
     private String organisationId;
+    private List<String> organisationProfileIds;
```
benouaer: I thought we're just persisting one profileId for the organisation

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/confirmeviction/ConfirmEviction.java
```diff
@@ -41,6 +41,7 @@ void configure(DecentralisedConfigBuilder<PCSCase, State, UserRole> configBuilde
                 .forState(State.CASE_ISSUED)
                 .name("Confirm the eviction details")
                 .grant(Permission.CRUD, UserRole.PCS_SOLICITOR)
+                .grant(Permission.CRUD, UserRole.GA_CLAIMANT_SOLICITOR)
```
benouaer, line 44: let's remove the UserRole.PCS_SOLICITOR wherever we add UserRole.GA_CLAIMANT_SOLICITOR
we can just name UserRole.CLAIMANT_SOLICITOR, since the GA was just added temporarily while we were still working on things so that we don't clash with existing roles

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/CreatePossessionClaim.java
```diff
@@ -93,27 +87,11 @@ private SubmitResponse<State> submit(EventPayload<PCSCase, State> eventPayload)
         PCSCase caseData = eventPayload.caseData();
 
         pcsCaseService.createCase(caseReference, caseData.getPropertyAddress(),
-                                  caseData.getLegislativeCountry());
-
-        String userId = securityContextService.getCurrentUserDetails().getUid();
-        scheduleRoleAssignment(caseReference, userId);
+                                  caseData.getLegislativeCountry(),
+                                  organisationService.getOrganisationIdForCurrentUser(),
+                                  organisationService.getOrgProfileIdsForCurrentUser());
```
benouaer: we don't use org ID and profile ID here, so we can fetch them where we use them

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/ResumePossessionClaim.java
```diff
@@ -86,6 +88,9 @@ public void configureDecentralised(DecentralisedConfigBuilder<PCSCase, State, Us
                 // The organisations that are the claimant themselves - local authority and the
                 // "other" profiles - hold this capacity rather than claimant-solicitor
                 .grant(Permission.CRUD, UserRole.CLAIMANT)
+                // The organisation's capacity covers the draft; CREATOR is the fallback for a
+                // case that derives no CaseAccessGroups.
+                .grant(Permission.CRUD, UserRole.CREATOR)
```
benouaer: I don't think we need this, we need claimant and claimant-solicitor

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/ResumePossessionClaim.java
```diff
@@ -187,6 +192,8 @@ public SubmitResponse<State> submitClaim(long caseReference, PCSCase pcsCase) {
         String organisationIdForCurrentUser = organisationService.getOrganisationIdForCurrentUser();
         pcsCaseService.createMainClaimOnCase(caseReference, pcsCase, organisationIdForCurrentUser);
 
+        scheduleCreatorRoleRevocation(caseReference);
```
benouaer: why are we moving this? we can just reuse the existing task

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/CreatePossessionClaim.java
```diff
@@ -93,27 +87,11 @@ private SubmitResponse<State> submit(EventPayload<PCSCase, State> eventPayload)
...
 
         return SubmitResponse.defaultResponse();
     }
 
-    private void scheduleRoleAssignment(long caseReference, String userId) {
```
benouaer, line 104: I think we can reuse this method

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/TestCaseGeneration.java
```diff
@@ -69,6 +71,10 @@ void configure(DecentralisedConfigBuilder<PCSCase, State, UserRole> configBuilde
                 .initialState(AWAITING_SUBMISSION_TO_HMCTS)
                 .showSummary()
                 .name(EVENT_NAME)
+                // Same claimant capacities as createPossessionClaim, which this mirrors.
+                .grant(Permission.CRUD, UserRole.GA_CLAIMANT_SOLICITOR)
+                .grant(Permission.CRUD, UserRole.CLAIMANT)
+                // Temporary: keeps the tool usable on AAT until the group access flags are on.
                 .grant(Permission.CRUD, UserRole.PCS_SOLICITOR)
```
benouaer, line 76: let's remove pcs_solicitor

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/DraftCaseDataRepository.java
```diff
@@ -9,22 +9,38 @@
 
 public interface DraftCaseDataRepository extends JpaRepository<DraftCaseDataEntity, Integer> {
 
-    Optional<DraftCaseDataEntity> findByCaseReferenceAndEventIdAndIdamUserId(
```
benouaer, line 12: I thought we were leaving the userId methods untouched so we don't break existing functionality, and adding methods for organisationId, and we use the relevant one based on scenario

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/PcsCaseRepository.java
```diff
@@ -12,4 +16,21 @@ public interface PcsCaseRepository extends JpaRepository<PcsCaseEntity, UUID> {
...
+    @Query("""
+        select distinct party.organisationId from PcsCaseEntity pcsCase
+        join pcsCase.parties party
+        where pcsCase.caseReference = :caseReference
+          and party.organisationId is not null
+          and (party.claimParties is empty
+               or exists (select claimParty from ClaimPartyEntity claimParty
+                          where claimParty.party = party and claimParty.role = :claimantRole))
+        """)
+    List<String> findOwningOrganisationIds(@Param("caseReference") long caseReference,
```
benouaer: not sure what this is used for?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
```diff
@@ -36,11 +40,18 @@ public class PartyService {
 
     private final PartyRepository partyRepository;
     private final AddressMapper addressMapper;
+    private final OrganisationService organisationService;
 
     public void createAllParties(PCSCase pcsCase, PcsCaseEntity pcsCaseEntity, ClaimEntity claimEntity,
                                  String organisationIdForCurrentUser) {
-        PartyEntity claimant = createClaimant(pcsCase, organisationIdForCurrentUser);
-        pcsCaseEntity.addParty(claimant);
+        List<String> orgProfileIds = organisationService.getOrgProfileIdsForCurrentUser();
```
benouaer: I thought we were filtering out the generic organisation profile, and just saving one organisation profile id
actually, we can probably do that filtering in the organisationService

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
```diff
@@ -141,15 +152,55 @@ public PartyEntity getPartyEntityById(UUID partyId, long caseReference) {
...
+                "Organisation profile IDs must be provided to create a case for organisation " + organisationId);
+        }
+        PartyEntity claimantParty = new PartyEntity();
+        claimantParty.setOrganisationId(organisationId);
+        claimantParty.setOrganisationProfileIds(organisationProfileIds);
+        pcsCaseEntity.addParty(claimantParty);
+    }
+
+    /** An organisation with no claim link is the party the case was created with, not a new one. */
+    private Optional<PartyEntity> findInitialisedClaimant(PcsCaseEntity pcsCaseEntity) {
```
benouaer: do we need this method? we can just check if there is a claimant party on the case.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
```diff
@@ -36,11 +40,18 @@ public class PartyService {
...
-        PartyEntity claimant = createClaimant(pcsCase, organisationIdForCurrentUser);
-        pcsCaseEntity.addParty(claimant);
+        List<String> orgProfileIds = organisationService.getOrgProfileIdsForCurrentUser();
+        PartyEntity claimant = findInitialisedClaimant(pcsCaseEntity)
+            .orElseGet(() -> {
+                PartyEntity created = new PartyEntity();
+                pcsCaseEntity.addParty(created);
+                return created;
+            });
+        populateClaimant(claimant, pcsCase, organisationIdForCurrentUser, orgProfileIds);
```
benouaer: we can leave pcsCaseEntity.addParty(created); outside the lambda, similar to how it was originally
so we get/create the claimant partyEntity, populate it and then add it

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
```diff
@@ -141,15 +152,55 @@ public PartyEntity getPartyEntityById(UUID partyId, long caseReference) {
...
+            .filter(party -> party.getClaimParties().isEmpty())
+            .findFirst();
+    }
+
+    /**
+     * Keeps what was validated at creation. rd-professional returns null rather than failing, so
+     * assigning unconditionally would wipe the organisation on a transient blip - and the same submit
+     * revokes CREATOR, leaving a case nobody can open.
+     */
+    private void setClaimantOrganisation(PartyEntity claimantParty, String organisationId,
```
benouaer: this method is declared before the method that uses it

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftCaseDataService.java
```diff
@@ -45,20 +107,65 @@ private UUID getCurrentUserId() {
...
+            .map(organisationId -> draftCaseDataRepository
+                .findByCaseReferenceAndEventIdAndOrganisationId(caseReference, eventId, organisationId))
+            .orElseGet(() -> draftCaseDataRepository
+                .findByCaseReferenceAndEventIdAndIdamUserIdAndPartyIdIsNullAndOrganisationIdIsNull(
+                    caseReference, eventId, userId));
+    }
+
     public Optional<PCSCase> getUnsubmittedCaseData(long caseReference, EventId eventId) {
         UUID userId = getCurrentUserId();
+        Optional<String> ownerOrganisationId = ownerOrganisationId(caseReference, eventId);
```
benouaer: can't we just check based on both org Id and userId?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftCaseDataService.java
```diff
@@ -10,32 +10,94 @@
...
 public class DraftCaseDataService {
 
+    /** HDPI-6221 keys drafts on the user so a citizen and their legal rep do not share one. */
+    enum DraftOwnership {
+        PARTY,
+        USER,
+        NO_DRAFTS
+    }
+
+    private static final Map<EventId, DraftOwnership> OWNERSHIP = new EnumMap<>(Map.ofEntries(
```
benouaer: do we need this ownership structure? can't we just search for for drafts based on both org and userId, or check if the user has an org then we search and save based on org, otherwise we use userId

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
```diff
@@ -51,6 +52,8 @@ public PcsCaseEntity createCase(long caseReference, AddressUK propertyAddress,
         pcsCaseEntity.setPropertyAddress(addressMapper.toAddressEntityAndNormalise(propertyAddress));
         pcsCaseEntity.setLegislativeCountry(legislativeCountry);
 
+        partyService.initialiseClaimant(pcsCaseEntity, organisationId, organisationProfileIds);
```
benouaer: initialiseClaimant is a bit generic. could we use a more specific name

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseRoleAssignmentTaskComponent.java
```diff
@@ -49,11 +49,13 @@ public CustomTask<RoleAssignmentTaskData> roleAssignmentTask() {
                 RoleAssignmentTaskData taskData = taskInstance.getData();
                 long caseReference = Long.parseLong(taskData.getCaseReference());
                 String userId = taskData.getUserId();
-                log.debug("Assigning claimant solicitor role and revoking creator role for case: {}", caseReference);
+                log.debug("Executing {} for case: {}", taskData.getAction(), caseReference);
 
                 try {
-                    caseRoleAssignmentService.assignRasRole(caseReference, userId, UserRole.CLAIMANT_SOLICITOR);
-                    caseRoleAssignmentService.revokeRasRole(caseReference, userId, UserRole.CREATOR);
+                    switch (taskData.getAction()) {
```
benouaer: why do we need the switch statement?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/CaseAccessGroupsUtil.java
```diff
@@ -0,0 +1,109 @@
...
+            .value(new CaseAccessGroup(CCD_ALL_CASES_ACCESS, groupId))
+            .build();
+    }
+
+    private static Optional<PartyRole> derivedRole(PartyEntity party) {
+        if (party.getOrganisationId() == null) {
+            return Optional.empty();
+        }
+        if (isClaimantCreatedWithTheCase(party)) {
+            return Optional.of(PartyRole.CLAIMANT).filter(DERIVED_ROLES::contains);
```
benouaer: this seems redundant

src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/CaseAccessGroupsUtil.java
```diff
@@ -0,0 +1,109 @@
...
+            .filter(DERIVED_ROLES::contains)
+            .findFirst();
+    }
+
+    /** No claim link until submit, so an organisation-bearing party without one is that claimant. */
+    private static boolean isClaimantCreatedWithTheCase(PartyEntity party) {
+        return party.getClaimParties().isEmpty();
+    }
+
+    private static String organisationProfileId(PartyEntity party) {
```
benouaer: I think we can move this logic to the organisation service, and only save one profile ID for each organisation, because in reality they only have one valid profile ID

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
```diff
@@ -141,15 +152,55 @@ public PartyEntity getPartyEntityById(UUID partyId, long caseReference) {
...
+                                   List<String> organisationProfileIds) {
+        Objects.requireNonNull(organisationId, "Organisation must be provided to create a case");
+        if (CollectionUtils.isEmpty(organisationProfileIds)) {
+            throw new IllegalArgumentException(
+                "Organisation profile IDs must be provided to create a case for organisation " + organisationId);
+        }
+        PartyEntity claimantParty = new PartyEntity();
+        claimantParty.setOrganisationId(organisationId);
+        claimantParty.setOrganisationProfileIds(organisationProfileIds);
+        pcsCaseEntity.addParty(claimantParty);
```
benouaer, line 184: we need to add the party role here as well, so that we can identify the party as claimant later

src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/CaseAccessGroupsUtil.java
```diff
@@ -0,0 +1,109 @@
...
+            .sorted()
+            .map(CaseAccessGroupsUtil::asStableListValue)
+            .toList();
+    }
+
+    /**
+     * Recomputed on every read over a HashSet of parties, so the id comes from the group rather than
+     * being random - otherwise an unchanged case differs read to read.
+     */
+    private static ListValue<CaseAccessGroup> asStableListValue(String groupId) {
```
benouaer: not sure what this is doing

src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/CaseAccessGroupsUtil.java
```diff
@@ -0,0 +1,109 @@
...
+    private static final Set<PartyRole> DERIVED_ROLES = Set.of(PartyRole.CLAIMANT);
+
+    private CaseAccessGroupsUtil() {
+    }
+
+    public static List<ListValue<CaseAccessGroup>> deriveCaseAccessGroups(Set<PartyEntity> parties) {
+        List<CaseAccessGroup> caseAccessGroups = new ArrayList<>();
+
+        parties.forEach(party -> derivedRole(party)
+            .flatMap(role -> GroupAccessType.forProfileAndRole(organisationProfileId(party), role))
```
benouaer: should GroupAccessType.forProfileAndRole return the caseAccessGroupTemplate instead of accessType

src/main/java/uk/gov/hmcts/reform/pcs/ccd/CaseType.java
```diff
@@ -108,6 +111,9 @@ public void configure(final ConfigBuilder<PCSCase, State, AccessProfile> builder
         builder.caseType(getCaseType(), getCaseTypeName(), CASE_TYPE_DESCRIPTION);
         builder.jurisdiction(JURISDICTION_ID, JURISDICTION_NAME, JURISDICTION_DESCRIPTION);
         builder.hmctsServiceId(hmctsServiceId);
+        // A group-less case is reachable only through the creator's case role, so it must pass
+        // case-type authorisation by itself - bracketed roles are otherwise skipped
+        builder.grantCaseTypeAccessToCaseRoles(AccessProfile.CREATOR);
```
benouaer: is this needed?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/CaseType.java
```diff
@@ -301,13 +309,7 @@ private void buildCaseDetailsTab(ConfigBuilder<PCSCase, State, AccessProfile> bu
                 "detailsTab_RequiredDocumentsDetails!=\"\"",
                 "## Required Documents"
             )
-            .field("detailsTab_RequiredDocumentsDetails")
-            .label(
-                "Documents you've uploaded",
-                "detailsTab_UploadedDocumentsChecklistDetails!=\"\"",
-                "## Documents you've uploaded"
-            )
-            .field("detailsTab_UploadedDocumentsChecklistDetails");
```
benouaer, line 313: why are we removing this? this doesn't seem to be related to GroupAccess

src/main/java/uk/gov/hmcts/reform/pcs/ccd/PCSCaseView.java
```diff
@@ -131,6 +136,14 @@ public PCSCase getCase(CaseViewRequest<State> request) {
         return pcsCase;
     }
 
+    private void applyCaseAccessGroups(PCSCase pcsCase, PcsCaseEntity pcsCaseEntity) {
```
benouaer: is this method called from anywhere else? if not, it doesn't seem to add value

src/main/java/uk/gov/hmcts/reform/pcs/reference/service/OrganisationService.java
```diff
@@ -76,6 +80,33 @@ public String getOrganisationIdForCurrentUser() {
...
     }
 
+    /**
+     * Retrieves the organisation identifier and organisation profile for the current user in a
+     * single rd-professional call. The profile is the one PRM keys the group access catalogue on;
+     * every organisation also carries the generic ORGANISATION_PROFILE, which is skipped.
+     *
+     * @return The organisation summary, or null if it cannot be resolved
+     */
+    public List<String> getOrgProfileIdsForCurrentUser() {
```
benouaer: should have a method to return org details for current user instead?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/AccessGrants.java
```diff
@@ -24,6 +25,7 @@ final class AccessGrants {
 
     static final UserRole[] PARTY_VISIBLE_ROLES = {
         CITIZEN,
+        CREATOR,
```
benouaer: do we need CREATOR role here?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/confirmeviction/ConfirmEviction.java
arun-hmcts: We'll do this in the clean-up ticket rather than now — removing it here affects other functionality.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/TestCaseGeneration.java
arun-hmcts: Same here — clean-up ticket rather than now, removing it affects other functionality.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/GroupAccessType.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/GroupAccessFields.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/GroupAccessType.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/PCSCase.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/party/PartyEntity.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/CreatePossessionClaim.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/ResumePossessionClaim.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/ResumePossessionClaim.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/CreatePossessionClaim.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/DraftCaseDataRepository.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/PcsCaseRepository.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftCaseDataService.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftCaseDataService.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseRoleAssignmentTaskComponent.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/CaseAccessGroupsUtil.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/CaseAccessGroupsUtil.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/CaseAccessGroupsUtil.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/CaseAccessGroupsUtil.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/CaseType.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/CaseType.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/PCSCaseView.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/AccessGrants.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
arun-hmcts: Role lives on claim_party and there is no claim yet at this point, so nothing to put it on. We set it at submit on the existing line `claimEntity.addParty(claimant, PartyRole.CLAIMANT)`. Before submit the organisation on the party is what tells us it is the claimant, only the claimant gets one.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
arun-hmcts: Yes we need it. Before submit there is no claim_party row to check, so the claimant is the party with the organisation. Without this we will create a second claimant next to the one created at case creation.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/reference/service/OrganisationService.java
arun-hmcts: DONE

config/owasp/suppressions.xml
```diff
@@ -43,6 +43,13 @@
     <cve>CVE-2026-63136</cve>
     <cve>CVE-2026-63140</cve>
   </suppress>
+  <suppress until="2026-08-23">
+    <notes><![CDATA[elasticsearch-rest-client-8.18.8.jar]]></notes>
+    <cve>CVE-2026-56144</cve>
+    <cve>CVE-2026-56145</cve>
+    <cve>CVE-2026-63136</cve>
+    <cve>CVE-2026-63140</cve>
+  </suppress>
```
scottstewart-sl, line 62: These CVE suppressions for ES and HttpClient5 can be removed now as they have been fixed in master.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/GroupAccessType.java
```diff
@@ -102,6 +113,32 @@ public enum GroupAccessType implements CCDAccessGroup {
         this.caseAssignedRoleField = caseAssignedRoleField;
     }
 
+    private static final Map<Key, GroupAccessType> BY_PROFILE_AND_ROLE = buildIndex();
```
scottstewart-sl: Minor - I think this constant should be near the top of the class

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/GroupAccessType.java
```diff
@@ -102,6 +113,32 @@ public enum GroupAccessType implements CCDAccessGroup {
...
+    /**
+     * The group ID template for an organisation profile acting in a party role, empty where the
+     * combination has no access type. Keyed lookup, so selection does not depend on the order these
+     * constants are declared in.
+     */
+    public static Optional<String> caseAccessGroupIdTemplateFor(String organisationProfileId, PartyRole partyRole) {
+        return partyRole == null
+            ? Optional.empty()
+            : Optional.ofNullable(BY_PROFILE_AND_ROLE.get(new Key(organisationProfileId, partyRole)))
+                .map(GroupAccessType::getCaseAccessGroupIdTemplate);
```
scottstewart-sl: Minor - this is mixing null checks and Optional. I think it should be possible to achieve this with something like, (untested):

```
return Optional.ofNullable(partyRole)
    .map(role -> BY_PROFILE_AND_ROLE.get(new Key(organisationProfileId, partyRole)))
    .map(GroupAccessType::getCaseAccessGroupIdTemplate);
```

Or you could even split the `new Key(organisationProfileId, partyRole)` into its own `.map()` step, so something like:

```
return Optional.ofNullable(partyRole)
    .map(role -> new Key(organisationProfileId, partyRole))
    .map(BY_PROFILE_AND_ROLE::get)
    .map(GroupAccessType::getCaseAccessGroupIdTemplate);
```

src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/party/PartyEntity.java
```diff
@@ -124,4 +125,9 @@ public class PartyEntity {
     @Fetch(FetchMode.SUBSELECT)
     private List<CasePartyFlagEntity> defendantFlags = new ArrayList<>();
 
+    /** Only the claimant is ever given an organisation, so carrying one is what makes a party the claimant. */
```
scottstewart-sl: That's true at the moment, but I'm not sure it's safe to always assume that is the case. The safest way to determine whether the party is a claimant is to look at their role on the main claim.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
```diff
@@ -141,15 +148,43 @@ public PartyEntity getPartyEntityById(UUID partyId, long caseReference) {
             ));
     }
 
-    private PartyEntity createClaimant(PCSCase pcsCase, String organisationIdForCurrentUser) {
+    /**
+     * The claimant party a case is created with, so CaseAccessGroups derive during the draft phase.
```
scottstewart-sl: Minor - it might be slightly clearer to change this comment to "The claimant party a **shell** case is created with..."

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
scottstewart-sl: It might be more resilient to have an explicit flag on the Party to indicate which one created the claim, or even check for the CREATOR role at this stage, (which wouldn't have been removed yet if I understand correctly). I think I prefer the explicit flag on Party as it may be something that is useful to know later on in the case even when the CREATOR role has been removed.

In future, there could be multiple claimants on a party, or some of the other parties like mortgagees could start having org details added.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
```diff
@@ -170,8 +205,21 @@ private PartyEntity createClaimant(PCSCase pcsCase, String organisationIdForCurr
...
 
-        return claimantParty;
+    /**
+     * Keeps what was validated at creation. rd-professional returns null rather than failing, so
+     * assigning unconditionally would wipe the organisation on a transient blip, and group access is
+     * the only way in, so the case would be left with nobody able to open it.
+     */
+    private void setClaimantOrganisation(PartyEntity claimantParty, String organisationId,
+                                         String orgProfileId) {
+        if (organisationId != null) {
```
scottstewart-sl, line 231: If these params are null/blank then something has gone wrong, (as you suggest in the comment), so could you log a warning?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/CaseAccessGroupsUtil.java
```diff
@@ -0,0 +1,80 @@
...
+    public static final String CCD_ALL_CASES_ACCESS = "CCD:all-cases-access";
+    static final String ORG_IDENTIFIER_TEMPLATE = "$ORGID$";
+
+    /** Claimant only for now; the defendant side belongs to the notice-of-change work. */
+    private static final Set<PartyRole> DERIVED_ROLES = Set.of(PartyRole.CLAIMANT);
+
+    private CaseAccessGroupsUtil() {
+    }
+
+    public static List<ListValue<CaseAccessGroup>> deriveCaseAccessGroups(Set<PartyEntity> parties) {
```
scottstewart-sl, line 32: It would be better to make this class a bean, without static method, so it can injected and more easily mocked for unit tests of the classes that use it.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/CaseAccessGroupsUtil.java
```diff
@@ -0,0 +1,80 @@
...
+            .id(UUID.nameUUIDFromBytes(groupId.getBytes(StandardCharsets.UTF_8)).toString())
+            .value(new CaseAccessGroup(CCD_ALL_CASES_ACCESS, groupId))
+            .build();
+    }
+
+    private static Optional<PartyRole> derivedRole(PartyEntity party) {
+        if (!party.hasOrganisation()) {
+            return Optional.empty();
+        }
+        if (isClaimantCreatedWithTheCase(party)) {
```
scottstewart-sl: If I understand this correctly, it's checking if there are no claim parties yet and if so, assuming that the party must be the claimant. That's fine, but the method name is talking about a specific step that will happen later on isn't it, (when the `resumePossessionClaim` event is submitted)?

If that is correct, then perhaps this method name would be more descriptive as something like `claimPartiesNotCreatedYet`

src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/CaseAccessGroupsUtil.java
arun-hmcts: This one is a pure function — parties in, groups out, no state and nothing injected. Making it a bean only helps to stub it, and stubbing it in PCSCaseViewTest would just prove we called it rather than that the group IDs are right. On something access related I would rather the real thing runs. There was no coverage there though, so I have added a test for it instead.

config/owasp/suppressions.xml
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/GroupAccessType.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/GroupAccessType.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/party/PartyEntity.java
arun-hmcts: DONE — hasOrganisation has gone, the claim creator flag on the party decides it now.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
arun-hmcts: DONE — the party now carries an explicit claim creator flag set at case creation, so we are not inferring the claimant from the organisation any more.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
arun-hmcts: DONE

src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/CaseAccessGroupsUtil.java
arun-hmcts: This check has gone now — the claimant is marked explicitly on the party, so there is nothing to infer.

scottstewart-sl (approved):

libanAbdirahman1 (approved):

benouaer (approved):

m014389b (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/TestCaseGeneration.java
benouaer: we need to remove it to make sure that things are running using the claimant-solicitor role rather than still relying on caseworker_pcs_solicitor

PR #2244, opened 2026-07-23
Hdpi 7594 pcs api logging restriction

tvr-hmcts:
https://tools.hmcts.net/jira/browse/HDPI-7594

Identified through investigating other issues within different environments.   The ticket HDPI-7594 gives an example of where PCS-API is leaking Personal Identifiable Information.  There are also other security concerns though from over exposure of knowledge beyond the application boundary.

This ticket does a few things to prevent that with the consideration over what is needed in order to diagnose issues within an environment at runtime.  There is another two tickets associated.  (one of which has the tests which cause breaks in the build when it is identified that a developer has carried out something by mistake).

Example (unless configured to do otherwise) logging of exception will be redacted to an Error Code taxonomy and output like:   "REDACTED [ErrorCode]".  This will allow some traceability before turning on greater diagnostics.

**The following are controlled by two configurations.**

1) A global override switch which can be set to true in other environments - SHOW_FULL_LOGS  I may need to turn this into a LaunchDarkly switch instead.

2) Normal logging configuration of a class setting to debug.

**There is a Redaction 'framework' is applied.**   

1) This can be seen in the addition of two 'base' exceptions - RedactedException and RedactedRuntimeException.  These delegate to RedactionGate.  

https://github.com/hmcts/pcs-api/blob/9fda528d4111071cc855e6c754529dbd1088473a/src/main/java/uk/gov/hmcts/reform/pcs/exception/RedactionGate.java

2) There is a SanitisingErrorDecoder which handles the Feign level exceptions and redacts them also.
 
https://github.com/hmcts/pcs-api/blob/9fda528d4111071cc855e6c754529dbd1088473a/src/main/java/uk/gov/hmcts/reform/pcs/config/SanitisingErrorDecoder.java
 
3) There is a RedactingThrowableConverter which is configured through logback-spring.xml. This redacts exception stack traces unless configured to do otherwise:

https://github.com/hmcts/pcs-api/blob/9fda528d4111071cc855e6c754529dbd1088473a/src/main/java/uk/gov/hmcts/reform/pcs/exception/RedactingThrowableConverter.java

4) ArchTest
Applying ArchTest to ensure that all custom exceptions extend from the RedactedRuntimeException
The ArchTest has some others in place also to align with the current design/architecture guidelines.

https://github.com/hmcts/pcs-api/blob/9fda528d4111071cc855e6c754529dbd1088473a/src/test/java/uk/gov/hmcts/reform/pcs/arch/ArchitectureTest.java

5) LoggerGate
Listens on changes to the logging level as a shortcut check

https://github.com/hmcts/pcs-api/blob/9fda528d4111071cc855e6c754529dbd1088473a/src/main/java/uk/gov/hmcts/reform/pcs/exception/LoggerGate.java


src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/genapp/SubmitEventHandler.java
```diff
@@ -194,7 +195,7 @@ private void validateCurrentUserIsLegalRepForParty(UUID currentUserId, UUID repr
             .isLegalRepresentativeLinkedToPartyAndActive(currentUserId, representedPartyId);
 
         if (!isLegalRepForParty) {
-            throw new PartyNotFoundException("No matching party found represented by current user");
+            throw new PartyNotFoundException(PARTY_NOT_FOUND);
```
scottstewart-sl, line 206: I'm a bit puzzled by some of the changes - where's the PII in this exception message?

I'd like to read this PR in more detail to understand what it is trying to solve.

tvr-hmcts: Yep there is no PII in this message.  However that is not the issue.   

There are two parts to the issue:

1) Allowing String's etc to be part of the construction of an exception or to be logged out in the via logger.[level] etc means things can leak - that is dangerous and so this becomes controlled now.  It is still possible to do with this, it just becomes controlled via the RedactionContext instead which is then managed by the ExpcetionRedaction

2) Although this does not leak PII - it can lead application boundary knowledge.   "I did X and I got Y"   That is a security issue.

The ErrorCode taxonomy can be tighter as the output here would still be PARTY_NOT_FOUND - which negates the second point a bit.

tvr-hmcts: The idea is so put in preventions.   Then have control over how things are seen within a production (or other exposed like demo) environment.

In the descriptions it says how this can work in other environments.

src/main/java/uk/gov/hmcts/reform/pcs/controllers/RestExceptionHandler.java
```diff
@@ -62,7 +61,8 @@ public ResponseEntity<Error> handleInvalidAuth(InvalidAuthTokenException ex) {
 
     @ExceptionHandler(CaseAccessException.class)
     public ResponseEntity<Error> handleCaseAccess(CaseAccessException ex) {
-        log.error("Case access denied", ex);
+        String message = "No defendants associated with this case";
+        log.error(message, ex);
         return ResponseEntity.status(HttpStatus.FORBIDDEN).body(new Error(ex.getMessage()));
     }
```
andrewdami11, line 76: REST callers may get a redacted message from `ex.getMessage()` here, but CCD callbacks get a friendly message from `CcdCallbackExceptionHandler`. Is it worth using the same user-facing message mapping in both places?

src/test/java/uk/gov/hmcts/reform/pcs/controllers/RestExceptionHandlerTest.java
```diff
@@ -339,40 +347,6 @@ void shouldHandleCaseAccessExceptionWithCause() {
         assertThat(responseEntity.getBody().message()).isEqualTo(expectedErrorMessage);
```
andrewdami11: This test expects a friendly message, but that only works because full exception messages are turned on in `@BeforeEach`. Would it be worth adding a test for the redacted production behaviour too, like `shouldHandleCaseAccessException`?

src/test/java/uk/gov/hmcts/reform/pcs/arch/ArchitectureTest.java
```diff
@@ -0,0 +1,117 @@
...
+    static final ArchRule controllers_should_not_expose_entities_via_return_type =
+        noMethods()
+            .that().areDeclaredInClassesThat().areAnnotatedWith(
+                org.springframework.web.bind.annotation.RestController.class)
+            .should().haveRawReturnType(resideInAPackage("..ccd.entity.."));
+
+    @ArchTest
+    static final ArchRule exceptions_must_extend_redacted_base =
+        priority(HIGH).classes()
+            .that().resideInAPackage("..exception..")
```
andrewdami11: This ArchUnit rule only checks exceptions under `..exception..`, but some redacted exceptions live in other packages (`BulkPrintMergeException`, `DocAssemblyException`, `NotificationException`, etc.). Is it worth applying the rule to all custom throwables under `uk.gov.hmcts.reform.pcs`?

src/main/java/uk/gov/hmcts/reform/pcs/exception/RedactingThrowableConverter.java
```diff
@@ -0,0 +1,38 @@
...
+    @Override
+    public String convert(ILoggingEvent event) {
+        if (event.getThrowableProxy() == null) {
+            return EMPTY_STRING;
+        }
+        // Global override (LOG_SHOW_FULL_EXCEPTIONS) OR DEBUG on the emitting logger -> full trace
+        if (SHOW_FULL_EXCEPTIONS || isDebugEnabled(event)) {
+            return super.convert(event);
+        }
+        return REDACTED + LINE_SEPARATOR;
```
andrewdami11, line 30: Redacted logs here only show `REDACTED`, with no error code. The PR description mentions `REDACTED [ErrorCode]`. Is it worth including the exception type and redacted message from `ExceptionRedaction.safeMessage()`, so logs are easier to trace?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
```diff
@@ -119,9 +132,9 @@ public PartyRole getPartyRole(PartyEntity partyEntity) {
...
                 .findFirst()
                 .map(ClaimPartyEntity::getRole)
-                .orElseThrow(() -> new PartyNotFoundException("Party not found on main claim"));
+                .orElseThrow(() -> new PartyNotFoundException(PARTY_NOT_FOUND));
     }
-    
+
     public PartyEntity getPartyEntityById(UUID partyId, long caseReference) {
         return partyRepository.findByIdAndPcsCaseCaseReference(partyId, caseReference)
             .orElseThrow(() -> new IllegalStateException(
```
andrewdami11: This `IllegalStateException` still includes party ID and case reference in the message, and `handleConflict` can return that text to clients. Is it worth moving this to `RedactedRuntimeException`?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/AccessCodeGenerationComponent.java
```diff
@@ -93,8 +93,8 @@ public CustomTask<AccessCodeTaskData> accessCodeGenerationTask() {
                     if (finalAttempt) {
                         MDC.put(MDC_TERMINAL_FAILURE, "true");
                         MDC.put(MDC_FAILURE_REASON, String.valueOf(e.getMessage()));
-                        log.error("Access code generation permanently failed for case {} party {} after {} "
-                                  + "attempts: {}", caseReference, defendantPartyId, attempt, e.getMessage(), e);
+                        log.error("Access code generation permanently failed for case {} party {} after {} ",
```
andrewdami11, line 96: This PR redacts exception output, but this log line still includes case reference and party ID. Has this been missed out?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/createpossessionclaim/CrossBorderPostcodeSelection.java
```diff
@@ -118,12 +120,12 @@ private AboutToStartOrSubmitResponse<PCSCase, State> midEvent(CaseDetails<PCSCas
             default -> {
                 //TODO: HDPI-1838 will handle multiple matches
                 throw new EligibilityCheckException(
-                    String.format(
+                    UNEXPECTED_ELIGIBILITY,
+                    RedactionContext.of("", String.format(
```
andrewdami11: `RedactionContext.of("", ...)` uses an empty key here, so debug output may look odd. Is it worth using a clear key like `"detail"` instead?

src/main/java/uk/gov/hmcts/reform/pcs/config/SanitisingErrorDecoder.java
```diff
@@ -0,0 +1,38 @@
...
+import uk.gov.hmcts.reform.pcs.exception.RedactionContext;
+import uk.gov.hmcts.reform.pcs.exception.RemoteCallException;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+
+public class SanitisingErrorDecoder implements ErrorDecoder {
+
+    @Override
```
andrewdami11, line 19: `SanitisingErrorDecoder` wraps Feign failures as `RemoteCallException`, but I could not see a unit test for it. Would it be worth adding a small test for status, method key, and body truncation?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/page/createpossessionclaim/CrossBorderPostcodeSelection.java
tvr-hmcts: Must have got distracted.   The key is missing.   Will apply once I check with master as to what it should be again.

src/main/java/uk/gov/hmcts/reform/pcs/exception/RedactingThrowableConverter.java
tvr-hmcts: These ones are slightly different.   This is when an exception is placed at the end of a logger statement.  I'll have a think on the other part though.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
tvr-hmcts: Interesting as it is not a pcs custom exception.   I will change over and as this does give up to much information.  I will scan for others.

src/test/java/uk/gov/hmcts/reform/pcs/arch/ArchitectureTest.java
tvr-hmcts: Those you've listed I actually caught using a different testing technique which I've not applied to this ticket.

However, I agree, I will adjust this rule and I am adding another for the different scenario you mentioned with IllegalStateException which currently falls through the cracks.

src/main/java/uk/gov/hmcts/reform/pcs/config/SanitisingErrorDecoder.java
tvr-hmcts: Ah missed.   Applying now thanks.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/AccessCodeGenerationComponent.java
tvr-hmcts: Should be through - RedactingMessageConverter -> [https://github.com/hmcts/pcs-api/blob/f975eea39b02302d22b8aec6e0a5fb25716068f2/src/main/java/uk/gov/hmcts/reform/pcs/exception/RedactingMessageConverter.java](url)

Reminds me though that I need to see how this is applied in the environments as this may require alternate configuration.

src/main/java/uk/gov/hmcts/reform/pcs/exception/RedactingThrowableConverter.java
tvr-hmcts: Although the touch point on this one is slightly different it maybe possible to get hold of the exception and then augment the output depending on the type so that any RedactedRuntimeException is output accordingly.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/party/PartyService.java
tvr-hmcts: Coming back to this.

Although I changed this it is part of the issue that is covered through the RedactingThrowableConverter / RedactingMessageConverter so long as they are applied within the logging setup.

src/main/java/uk/gov/hmcts/reform/pcs/exception/RedactingThrowableConverter.java
tvr-hmcts: Updated for when it is a RedactedRuntimeException

[https://github.com/hmcts/pcs-api/blob/b0958032b45c23d0be2b0ebe7c8ca368f09f4975/src/main/java/uk/gov/hmcts/reform/pcs/exception/RedactingThrowableConverter.java](url)

src/test/java/uk/gov/hmcts/reform/pcs/controllers/RestExceptionHandlerTest.java
```diff
@@ -339,53 +387,20 @@ void shouldHandleCaseAccessExceptionWithCause() {
...
     void shouldMapIdamExceptionWrappingOAuth2WithRestClient429ToServiceUnavailable() {
         // Real production shape: Spring's OAuth2 password client wraps a RestClient 429 in
         // OAuth2AuthorizationException, then SystemUpdateUserTokenProvider wraps that in IdamException.
         HttpClientErrorException tooMany = HttpClientErrorException.create(
-            HttpStatus.TOO_MANY_REQUESTS, "Too Many Requests", HttpHeaders.EMPTY, new byte[0], null);
+            TOO_MANY_REQUESTS, "Too Many Requests", HttpHeaders.EMPTY, new byte[0], null);
         OAuth2Error oauthError = new OAuth2Error("invalid_token_response", "throttled by IDAM", null);
         OAuth2AuthorizationException oauthEx = new OAuth2AuthorizationException(oauthError, tooMany);
-        IdamException ex = new IdamException("Unable to get access token response", oauthEx);
+        IdamException ex = new IdamException(AUTH_TOKEN_EMPTY,
```
andrewdami11: These 503 cases use `AUTH_TOKEN_EMPTY`, which always counts as upstream-unavailable, so do they ever exercise the OAuth2 or `statusCode` paths?
Production token errors use `AUTH_TOKEN_RETRIEVAL_FAIL`, and the Feign 4xx case below may put status under `OAUTH2_ERROR_CODE` instead of `AUTH_VALIDATION_STATUS_CODE`. Would it be worth aligning our ones in that fashion?

src/main/java/uk/gov/hmcts/reform/pcs/exception/RedactingMessageConverter.java
```diff
@@ -0,0 +1,24 @@
...
+
+public final class RedactingMessageConverter extends ClassicConverter {
+
+    @Override
+    public String convert(ILoggingEvent event) {
+        Object[] arguments = event.getArgumentArray();
+        if (arguments == null || arguments.length == 0) {
+            return event.getMessage();
+        }
+        Object[] redactedArguments = new Object[arguments.length];
```
andrewdami11, line 24: This redacts every `{}` argument, and unlike `RedactingThrowableConverter` it ignores `LOG_SHOW_FULL_EXCEPTIONS` and DEBUG. Is it worth honouring those same switches, or only redacting args that look sensitive, so ops logs stay useful?

src/main/java/uk/gov/hmcts/reform/pcs/config/SanitisingErrorDecoder.java
```diff
@@ -0,0 +1,38 @@
...
+    public Exception decode(String methodKey, Response response) {
+        return new RemoteCallException(ErrorCode.REMOTE_CALL,
+                                       RedactionContext.builder()
+                                           .value("Remote call", methodKey)
+                                           .value("Status", response.status())
+                                           .value("Response Body", safeReadBody(response))
```
andrewdami11: The body is truncated but still stored raw under `Response Body`, so `LOG_SHOW_FULL_EXCEPTIONS` can log Feign payloads with personal data or tokens. Would it be worth omitting the body from context, or hashing or redacting it, instead of keeping the raw text?

src/main/java/uk/gov/hmcts/reform/pcs/exception/RedactingThrowableConverter.java
```diff
@@ -0,0 +1,48 @@
...
+            return EMPTY_STRING;
+        }
+        // Global override (LOG_SHOW_FULL_EXCEPTIONS) OR DEBUG on the emitting logger -> full trace
+        if (SHOW_FULL_EXCEPTIONS || isDebugEnabled(event)) {
+            return super.convert(event);
+        }
+
+        if (proxy instanceof ThrowableProxy throwableProxy) {
+            Throwable source = throwableProxy.getThrowable();
+            if (source instanceof RedactedRuntimeException) {
```
andrewdami11, line 26: This new branch returns `getMessage()` for `RedactedRuntimeException`, but the converter tests still only cover a plain `IllegalStateException`. Should we add a case that asserts the redacted `REDACTED [ErrorCode]` form here?

src/test/java/uk/gov/hmcts/reform/pcs/controllers/RestExceptionHandlerTest.java
tvr-hmcts: I have adjusted the RestExceptionHandlerTest to handle both.

src/test/java/uk/gov/hmcts/reform/pcs/controllers/RestExceptionHandlerTest.java
tvr-hmcts: I think you may have hit the nail on the head there.

I applied it to align with what was currently in place.   Though when then trying to get the tests for the UpstreamThrottling aligned something did not seem correct.

I can take another look in a bit.

src/main/java/uk/gov/hmcts/reform/pcs/exception/RedactingMessageConverter.java
tvr-hmcts: Good catch.  Thank you.   Will apply.

src/main/java/uk/gov/hmcts/reform/pcs/config/SanitisingErrorDecoder.java
tvr-hmcts: Good thought.  I took it that if the debug or LOG_... was applied then there would be a lot of exposure anyway.   The ticket is about trying to squash those habits that are through out the codebase but balancing it with the need in production to be able to diagnose issues.   Perhaps here there can be something else I can grab instead.

I tried to avoid hashing or regx as they are a bit fragile.

src/main/resources/application.yaml
```diff
@@ -275,6 +275,7 @@ send-letter:
   url: ${SEND_LETTER_URL:http://localhost:8485}
 
 logging:
+  config: classpath:pcs-logback.xml
   level:
     uk.gov.hmcts.reform.pcs.ccd.service.bulkprint: INFO
```
guygrewal77, line 280: Should this line be here ? It only relates to Bulk Prints.

tvr-hmcts: Oh that one is in the master branch.   I left as is and just added the other line.

src/main/java/uk/gov/hmcts/reform/pcs/exception/RedactingThrowableConverter.java
tvr-hmcts: The testing was enhanced.   Can you let me know though if it covers what you mean.

https://github.com/hmcts/pcs-api/blob/3e960d38dc95731165f77f20fbffb7f14699832f/src/test/java/uk/gov/hmcts/reform/pcs/exception/RedactingThrowableConverterTest.java

src/test/java/uk/gov/hmcts/reform/pcs/controllers/RestExceptionHandlerTest.java
andrewdami11: Thanks — the AUTH_TOKEN_RETRIEVAL_FAIL / AUTH_VALIDATION fixtures look right now, and AUTH_TOKEN_EMPTY is kept only for the real empty-token path. Resolving.

src/main/java/uk/gov/hmcts/reform/pcs/exception/RedactingThrowableConverter.java
andrewdami11: Thanks — yes, PartyNotFoundException in the matrix covers the RedactedRuntimeException branch well enough. Resolving.

andrewdami11 (commented):
.

src/main/java/uk/gov/hmcts/reform/pcs/config/RemoteCallExceptionTranslationFilter.java
```diff
@@ -0,0 +1,46 @@
...
+    @Override
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
+        throws ServletException, IOException {
+        try {
+            filterChain.doFilter(request, response);
+        } catch (RemoteCallException ex) {
+            int downstream = ex.getStatus();   // status preserved by SanitisingErrorDecoder
+            int translated = downstream == HttpStatus.UNAUTHORIZED.value()
+                ? HttpStatus.UNAUTHORIZED.value()
+                : HttpStatus.INTERNAL_SERVER_ERROR.value();
```
andrewdami11, line 35: I could not see a unit test for this new filter. Would it be worth adding a couple of cases for 401 vs other statuses?

src/main/java/uk/gov/hmcts/reform/pcs/idam/IdamAuthenticator.java
```diff
@@ -19,25 +26,35 @@ public class IdamAuthenticator {
     public User validateAuthToken(String authorisation) {
         if (authorisation == null || authorisation.isBlank()) {
             log.warn("Authorization token is null or blank");
-            throw new InvalidAuthTokenException("Authorization token is null or blank");
+            throw new InvalidAuthTokenException(AUTH_BLANK);
         }
         if (!authorisation.startsWith(BEARER_PREFIX) || authorisation.length() <= 7) {
             log.warn("Malformed Bearer token: '{}'", authorisation);
-            throw new InvalidAuthTokenException("Malformed Authorization token");
+            throw new InvalidAuthTokenException(AUTH_MALFORMED);
```
andrewdami11, line 33: This warn log still prints the full Authorization header when the Bearer token is malformed. Is it worth redacting that value (or logging only the length), so a live token doesn't accidentally surface?

src/main/java/uk/gov/hmcts/reform/pcs/config/RemoteCallExceptionTranslationFilter.java
tvr-hmcts: Yep just need to push along with another existing security configuration class which had no tests.

src/main/java/uk/gov/hmcts/reform/pcs/idam/IdamAuthenticator.java
tvr-hmcts: Will change this ... however, it would only get logged out IF show all logs was turned on or the debugging level set for the class.

That is covered in the RedactingMessageConverter

src/main/java/uk/gov/hmcts/reform/pcs/config/RemoteCallExceptionTranslationFilter.java
andrewdami11: Thanks — `RemoteCallExceptionTranslationFilterTest` now covers pass-through, 401, and non-401 → 500. Resolving.

src/main/java/uk/gov/hmcts/reform/pcs/idam/IdamAuthenticator.java
andrewdami11: Thanks — the malformed Bearer path now logs length only, not the header value. Resolving.

andrewdami11 (dismissed):

scottstewart-sl (changes requested):
I think this needs further discussions before it can be merged

PR #2247, opened 2026-07-23
Dispose of inactive drafts after 30 days

banderous:
Adds disposal for draft cases inactive for more than 30 days.

  - Identifies inactive draft cases.
  - SDK scheduled task will move them to PendingDisposal.
  - Sets the CCD disposal TTL to trigger central disposal
  - Deletes locally held PCS case data after central disposal.

See [sdk documentation](https://github.com/hmcts/dtsse-ccd-config-generator/blob/master/docs/retain-and-dispose.md) for more details including dry runs & live operations.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/State.java
+++ b/src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/State.java
@@ -10,6 +10,7 @@ import uk.gov.hmcts.reform.pcs.ccd.accesscontrol.DefendantAccess;
 import uk.gov.hmcts.reform.pcs.ccd.accesscontrol.GlobalSearchAccess;
 import uk.gov.hmcts.reform.pcs.ccd.accesscontrol.InternalCaseFlagAccess;
 import uk.gov.hmcts.reform.pcs.ccd.accesscontrol.RasValidationAccess;
+import uk.gov.hmcts.reform.pcs.ccd.accesscontrol.SystemUserAccess;
 
 /**
  * All possible PCS case states.
@@ -104,6 +105,11 @@ public enum State {
             GlobalSearchAccess.class},
         hint = "${caseTitleMarkdown}"
     )
-    BREATHING_SPACE
-}
+    BREATHING_SPACE,
 
+    @CCD(
+        label = "Pending Disposal",
+        access = {SystemUserAccess.class}
+    )
+    PendingDisposal
+}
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/State.java
```diff
@@ -104,6 +105,11 @@ public enum State {
...
     )
-    BREATHING_SPACE
-}
+    BREATHING_SPACE,
 
+    @CCD(
+        label = "Pending Disposal",
+        access = {SystemUserAccess.class}
+    )
+    PendingDisposal
```
scottstewart-sl, line 114: Could we have this in upper snake case to match the other enum value here? Or if it has to have this exact name then could you add a comment in the code to say why it is different?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/RetainAndDispose.java
```diff
@@ -0,0 +1,37 @@
...
+        configBuilder
+            .event(DISPOSAL_EVENT_ID)
+            .forStateTransition(AWAITING_SUBMISSION_TO_HMCTS, PendingDisposal)
+            .name("Mark draft for disposal")
+            .description("Move an inactive draft into its disposal state")
+            .grant(Permission.CRU, SYSTEM_USER);
+
+        configBuilder
+            .event(CONFIRM_DISPOSAL_EVENT_ID)
+            .forStateTransition(PendingDisposal, PendingDisposal)
```
scottstewart-sl, line 31: Minor - is there a reason for specifying this as a transition rather than just having `.forState(PendingDisposal)` ?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/retention/PcsRetainAndDisposePolicy.java
```diff
@@ -0,0 +1,54 @@
...
+                "inactivityDays", DRAFT_INACTIVITY_DAYS
+            ),
+            Long.class
+        );
+    }
+
+    @Override
+    public void dispose(long caseReference) {
+        pcsCaseRepository.findByCaseReference(caseReference)
+            .ifPresent(pcsCaseRepository::delete);
```
scottstewart-sl, line 52: This would also need to delete from the `DraftCaseDataRepository`, which contains the save and resume answers before the user has submitted them in ExUI.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/State.java
guygrewal77: Ignore these changes @scottstewart-sl. I am implementing the delete case in a different PR. This PR is only a draft and not for wider review.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/RetainAndDispose.java
banderous: can do

PR #2250, opened 2026-07-23
HDPI-7643: Number defendants according to rank for View the Claim and View the Response pages

mazzopardi2:
### Jira link

See [HDPI-7643](https://tools.hmcts.net/jira/browse/HDPI-7643)

### Change description
- exposes each defendant's rank so that the citizen dashboard can number defendants consistently with the rank in use instead of hard-coded values

### Testing done
- added test and manually tested

  * [ ] Yes
  * [x] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/PartiesView.java
```diff
@@ -65,6 +65,9 @@ private ListValue<Party> toListValue(ClaimPartyEntity claimPartyEntity, boolean
         Party party = shouldRedact
             ? toPartialParty(partyEntity)
             : toParty(partyEntity);
+        if (claimPartyEntity.getRole() == PartyRole.DEFENDANT) {
+            party.setRank(claimPartyEntity.getRank());
+        }
```
scottstewart-sl: You might as well do it for all party roles, unless there is a specific reason to only do it for defendants?

mazzopardi2: Yep fair point - there's no reason to set the rank only for defendants . I've removed the role check now- thank you!

scottstewart-sl (approved):

src/test/java/uk/gov/hmcts/reform/pcs/ccd/view/PartiesViewTest.java
```diff
@@ -164,6 +164,26 @@ void shouldMapAllParties() {
             .containsExactly("Dave", "Eve");
     }
 
+    @Test
+    void shouldMapRankForDefendants() {
```
andrewdami11: Now that `rank` is mapped for every party role, could we also assert claimants (and underlessees if useful) get their ranks, or rename this test so it does not read as defendant-only?

mazzopardi2: Good spot thank you, renamed to shouldMapRankForAllPartyRoles and extended it to assert claimants, defendants and underlessees get their rank.

matthewsearle01 (approved):

PR #2277, opened 2026-07-27
Hdpi 3404 scheduled case deletion

guygrewal77:
### Jira link

See [HDPI-3404](https://tools.hmcts.net/jira/browse/HDPI-3404)

### Change description

This PR allows a scheduler to delete unissued cases that were created 30 days ago.

There are a number of steps performed to accomplish this using a scheduler - 

- Finds cases that are in 'AWAITING_SUBMISSION_TO_HMCTS' and 'PENDING_CASE_ISSUED' states and were created 30 days ago or more
- For each case, the markCaseForDeletion event is run to move the state to 'DRAFT_DISCARDED'
- The confirmCaseDisposal event is then run to set the ttlIncrement to -1
- The above causes the resolvedTTl to be set for the case in case_data to yesterday's date. The next ccd-case-disposer scheduled task which runs daily then picks up the case to be immediately removed from elastic search and the central ccd database
- The next step deletes data from the decentralised ccd schema and draft, PUBLIC schemas. This also includes deleting any documents from cdam.

Please note: The parameters will need to be changed once DA/QA is complete to have it run daily and with 30 days for draft discard. Also the preview-env will need the scheduler taken off.

### Testing done

Tested in preview environment by setting EXPIRED_CASE_DELETION_DISCARD_AFTER_DAYS to -1 and the EXPIRED_CASE_DELETION_SCHEDULE to every 5 minutes. This allows the above steps to run every 5 minutes to pick up cases eligible for deletion immediately. Verified in the database that the records get deleted and the case no longer shows in the XUI Case List.

  * [ ] Yes
  * [X] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/ClaimantAccess.java
```diff
@@ -15,6 +16,7 @@ public class ClaimantAccess implements HasAccessControl {
     public SetMultimap<HasRole, Permission> getGrants() {
         SetMultimap<HasRole, Permission> grants = HashMultimap.create();
         grants.putAll(PCS_SOLICITOR, Permission.CRU);
+        grants.putAll(SYSTEM_USER, Permission.CRU);
```
gmmagruder: I think it would be better for SYSTEM_USER to have it's own access control class. This way if there's any update to ClaimantAccess, it won't affect the system user's access.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/DraftDiscardedAccess.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.hmcts.reform.pcs.ccd.accesscontrol;
+
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.SetMultimap;
+import uk.gov.hmcts.ccd.sdk.api.HasAccessControl;
+import uk.gov.hmcts.ccd.sdk.api.HasRole;
+import uk.gov.hmcts.ccd.sdk.api.Permission;
+
+public class DraftDiscardedAccess implements HasAccessControl {
```
gmmagruder: Following from above - maybe rename this to SystemUserAccess? There will likely be other system events/cron jobs in the future that will also use the system user.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/PCSCase.java
```diff
@@ -746,4 +747,11 @@ public class PCSCase {
         typeParameterOverride = "CaseStateOption"
     )
     private CaseStateOption targetState;
+
+    @JsonProperty("TTL")
+    @CCD(
+            label = "Set up TTL",
+            typeOverride = FieldType.TTL
```
gmmagruder: Is the label required? This shouldn't be displayed in ExUI.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/DraftDiscardedAccess.java
guygrewal77: Addressed in latest.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/PCSCase.java
guygrewal77: Addressed in latest.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/ClaimantAccess.java
guygrewal77: Addressed in latest.

src/main/resources/application.yaml
```diff
@@ -269,6 +269,10 @@ bulk-print:
   # BULK_PRINT_LOOKBACK_HOURS (lower envs only): sweep ignores packs whose documents are older than N hours,
   # so enabling the flag cannot flood send-letter with the historical backlog. Unset in prod = full backlog.
 
+expired-case-deletion:
+  schedule: ${EXPIRED_CASE_DELETION_SCHEDULE:FIXED_DELAY|60s}
+  discard-after-days: ${EXPIRED_CASE_DELETION_DISCARD_AFTER_DAYS:-1}
```
andrewdami11: The defaults here are `FIXED_DELAY|60s` and `discard-after-days: -1`, and AAT/prod charts do not override them. Preview overrides look intentional for testing, but a merge would inherit these unsafe defaults in higher envs. Would it be worth defaulting to `30` days and a daily schedule in `application.yaml`, and keeping `-1` only in preview? Or perhaps anything that is more in line with the higher envs, vs the lower ones?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,74 @@
...
+    private void performCaseDeletionTasks(long caseRef) {
+        log.debug("Performing case deletion tasks for case: {}", caseRef);
+        Runnable caseDeletionTasks = () -> {
+            try {
+                ccdCaseDataDeletionService.markCaseForDeletion(caseRef);
+                ccdCaseDataDeletionService.confirmCaseDisposal(caseRef);
+            } catch (CaseNotFoundException e) {
+                log.error("Case not found in main ccd datastore. Will proceed to delete in decentralised ccd schema");
+            }
+            caseDeletionService.deleteCase(caseRef);
```
andrewdami11: `caseDeletionService.deleteCase` always runs after mark/confirm, even when those CCD calls fail quietly and return `null`. A mid-flight state change or transient CCD error could then hard-delete local data while central CCD stays live. Would it be worth only deleting locally when both events succeed, or when the case is confirmed missing?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/casedeletion/CcdCaseDataDeletionService.java
```diff
@@ -0,0 +1,94 @@
...
+                    String.valueOf(caseRef), submitContent);
+            log.debug("Completed event: {} for case: {}", eventId.name(), caseRef);
+            return caseResource;
+        } catch (FeignException e) {
+            if (e.getMessage().contains("Case ID is not valid")) {
+                throw new CaseNotFoundException(caseRef);
+            }
+            log.error("Error running event: {} for case: {}. Error: {}", eventId.name(), caseRef, e.getMessage());
+        }
+        return null;
```
andrewdami11: Unexpected `FeignException`s are logged here and then `null` is returned. Callers cannot tell success from failure, which makes it easy for the scheduler to keep deleting local data. Is it worth rethrowing unexpected `FeignException`s (or wrapping them) instead of returning `null`?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
```diff
@@ -93,6 +96,14 @@ public PcsCaseEntity loadCase(long caseReference) {
             .orElseThrow(() -> new CaseNotFoundException(caseReference));
     }
 
+    @Transactional
```
andrewdami11, line 102: CDAM document deletes run inside this `@Transactional` method (and under `CaseDeletionService.deleteCase`). That can hold a DB connection open during HTTP calls, and a CDAM failure or rollback can leave documents and rows out of sync. Would it be worth deleting CDAM documents outside the DB transaction, or after a successful commit?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftCaseDataService.java
```diff
@@ -333,6 +333,17 @@ public void deleteUnsubmittedCaseData(long caseReference,
         );
     }
 
+    @Transactional
+    public void deleteUnsubmittedCaseDataBySystemUser(long caseReference, EventId eventId) {
```
andrewdami11, line 326: This deletes a single `Optional` row for `resumePossessionClaim`, but the unique index is on `(case_reference, event_id, idam_user_id, party_id)`, so multiple draft rows can exist. I think that might risk `IncorrectResultSizeDataAccessException`, and other event drafts may be left behind. Is it worth deleting all draft rows for the case reference instead?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/casedeletion/CaseDeletionService.java
```diff
@@ -0,0 +1,28 @@
...
+
+    private final CcdCaseDataDeletionService ccdCaseDataDeletionService;
+    private final DraftCaseDataService draftCaseDataService;
+    private final PcsCaseService pcsCaseService;
+
+    @Transactional
+    public void deleteCase(long caseReference) {
+        ccdCaseDataDeletionService.deleteCcdCaseData(caseReference);
+        draftCaseDataService.deleteUnsubmittedCaseDataBySystemUser(caseReference, resumePossessionClaim);
+        pcsCaseService.deleteCase(caseReference);
```
andrewdami11: If mark/confirm succeed and then `pcsCaseService.deleteCase` throws `CaseNotFoundException`, the local transaction rolls back, but the case is already `DRAFT_DISCARDED` with TTL. The finder only selects unissued states, so it will not be retried. Would it be worth treating a missing PUBLIC `pcs_case` as a no-op, and/or adding a retry path for `DRAFT_DISCARDED`?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/CcdCaseRepository.java
```diff
@@ -0,0 +1,62 @@
...
+                DELETE FROM ccd.case_event
+                WHERE case_data_id IN (SELECT id FROM deleted_case_data)
+                RETURNING id
+            ),
+            deleted_case_event_audit AS (
+                DELETE FROM ccd.case_event_audit
+                WHERE case_event_id IN (SELECT id FROM deleted_case_event)
+                RETURNING case_event_id
+            )
+            DELETE FROM ccd.case_event_significant_items
```
andrewdami11: `case_event_significant_items` is deleted only for event IDs returned from `case_event_audit`. Events without audit rows may leave orphans, and `case_data` is deleted before its children in the chain. Would it be worth deleting significant items from `deleted_case_event` IDs directly, and deleting child tables before `case_data`?

src/test/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTaskTest.java
```diff
@@ -0,0 +1,110 @@
...
+        runSweepViaReflection();
+
+        // Then
+        verify(ccdCaseDataDeletionService).findExpiredDraftCases(discardAfterDays);
+        verifyNoMoreInteractions(ccdCaseDataDeletionService);
+        verifyNoMoreInteractions(caseDeletionService);
+    }
+
+    @Test
+    void shouldPerformCaseDeletionTasksForEachExpiredCase() throws Exception {
```
andrewdami11: Happy-path ordering is covered well here, but there are no tests for CCD event failure (`null`/exception), `CaseNotFoundException`, or abort-on-failure behaviour. Is it worth adding those paths so the scheduler continue-or-abort rules are locked down?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/CcdCaseRepository.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+
+@Repository
+@AllArgsConstructor
+public class CcdCaseRepository {
+
+    private final JdbcTemplate jdbcTemplate;
+    private final NamedParameterJdbcTemplate namedParameterJdbcTemplate;
+
+    public List<DraftCasesToDiscard> findExpiredDraftCases(int discardAfterDays) {
```
andrewdami11: `findExpiredDraftCases` has no batch limit. A large backlog could make one sweep run for a long time. Would it be worth limiting each sweep to a batch size, for example `LIMIT 100`?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/CcdCaseRepository.java
```diff
@@ -0,0 +1,62 @@
...
+                """
+                   SELECT cd.reference
+                   FROM ccd.case_data cd
+                   WHERE cd.created_date < now()::date - :discardDaysAfter
+                   AND cd.state in ('AWAITING_SUBMISSION_TO_HMCTS', 'PENDING_CASE_ISSUED')
+                """,
+                namedParameters,
+                (rs, rowNum) -> {
+                    long caseRef = rs.getLong("reference");
+                    return DraftCasesToDiscard.builder()
```
gmmagruder: Could a list of case references be returned instead? It seems redundant to create a class just to store the the case reference.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: There could be scenarios where there is still data in the decentralised ccd schema even though it's cleared from the central data store. The decentralised schemas are decoupled from the central ccd datastore.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/casedeletion/CcdCaseDataDeletionService.java
```diff
@@ -0,0 +1,94 @@
...
+
+            CaseResource caseResource = coreCaseDataApi.createEvent(idamToken, serviceAuthorization,
+                    String.valueOf(caseRef), submitContent);
+            log.debug("Completed event: {} for case: {}", eventId.name(), caseRef);
+            return caseResource;
+        } catch (FeignException e) {
+            if (e.getMessage().contains("Case ID is not valid")) {
+                throw new CaseNotFoundException(caseRef);
+            }
+            log.error("Error running event: {} for case: {}. Error: {}", eventId.name(), caseRef, e.getMessage());
```
tvr-hmcts: Pls remove the e.getMessage() and apply as an exception at the end of the log instead.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/CcdCaseRepository.java
guygrewal77: The idea is to process each case deletion in a separate thread.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/casedeletion/ConfirmCaseDisposal.java
```diff
@@ -0,0 +1,40 @@
...
+@AllArgsConstructor
+public class ConfirmCaseDisposal implements CCDConfig<PCSCase, State, UserRole> {
+
+    @Override
+    public void configureDecentralised(DecentralisedConfigBuilder<PCSCase, State, UserRole> configBuilder) {
+        configBuilder
+                .decentralisedEvent(confirmCaseDisposal.name(), this::submit)
+                .forStates(State.DRAFT_DISCARDED)
+                .name("Discard unissued cases")
+                .ttlIncrement(-1)
```
tvr-hmcts, line 30: What does setting it here do?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
tvr-hmcts: && || schedule a job from this point.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/casedeletion/CcdCaseDataDeletionService.java
```diff
@@ -0,0 +1,94 @@
...
+    public List<DraftCasesToDiscard> findExpiredDraftCases(int discardAfterDays) {
+        return ccdCaseRepository.findExpiredDraftCases(discardAfterDays);
+    }
+
+    public void deleteCcdCaseData(long caseReference) {
+        ccdCaseRepository.deleteCcdCaseData(caseReference);
+        log.debug("Deleted case data for case reference: {}", caseReference);
+    }
+
+    public CaseResource markCaseForDeletion(long caseRef) throws CaseNotFoundException {
```
tvr-hmcts: The CaseNotFoundException does not need to be declared

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/casedeletion/CcdCaseDataDeletionService.java
```diff
@@ -0,0 +1,94 @@
...
+        log.debug("Marking following case for deletion: {} ", caseRef);
+
+        return performEvent(markCaseForDeletion, caseRef);
+    }
+
+    public CaseResource confirmCaseDisposal(long caseRef) throws CaseNotFoundException {
```
tvr-hmcts: The CaseNotFoundException does not need to be declared

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,74 @@
...
+            }
+            caseReferences.forEach(this::performCaseDeletionTasks);
+        } finally {
+            MDC.remove(MDC_TASK_NAME);
+        }
+    }
+
+    private void performCaseDeletionTasks(long caseRef) {
+        log.debug("Performing case deletion tasks for case: {}", caseRef);
+        Runnable caseDeletionTasks = () -> {
```
tvr-hmcts: The runnable is not needed for this and it may cause problems in how the Task with DbScheduler handles things.

tvr-hmcts (commented):
Timeboxed thoughts.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/CcdCaseRepository.java
guygrewal77: The script deletes from the 'case_event_significant_items' table even if there isn't a value in 'case_event_audit'.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
guygrewal77: Moved the document deletion outside of the case deletion method.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftCaseDataService.java
guygrewal77: I was thinking that might be the case actually but thought there could only be 1 draft per case.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/casedeletion/CaseDeletionService.java
guygrewal77: This scenario should never happen as there will be data in the PUBLIC/draft schema once a case is created.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/casedeletion/ConfirmCaseDisposal.java
guygrewal77: The ttlIncrement of -1 sets the resolvedttl date in the case_data to yesterday to trigger the ccd-case-disposer to run the case deletion immediately. Before merging this will be set to 0.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/CcdCaseRepository.java
guygrewal77: Updated in latest

src/main/resources/application.yaml
guygrewal77: Updated in latest

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/casedeletion/CcdCaseDataDeletionService.java
guygrewal77: Addressed in latest.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/DraftCaseDataService.java
guygrewal77: Addressed in latest.

src/test/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTaskTest.java
guygrewal77: Addressed in latest.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/casedeletion/CcdCaseDataDeletionService.java
guygrewal77: Addressed in latest.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/casedeletion/CcdCaseDataDeletionService.java
guygrewal77: Addressed in latest.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/casedeletion/CcdCaseDataDeletionService.java
guygrewal77: Addressed in latest.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: Changed the approach to use multi threading which processes each case in parallel.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/casedeletion/CaseDeletionService.java
```diff
@@ -0,0 +1,58 @@
...
+@Service
+@AllArgsConstructor
+@Slf4j
+public class CaseDeletionService {
+
+    private final CcdCaseDataDeletionService ccdCaseDataDeletionService;
+    private final DraftCaseDataService draftCaseDataService;
+    private final PcsCaseService pcsCaseService;
+
+    @Transactional
```
andrewdami11: This whole method is `@Transactional`, and it still calls CCD Feign events plus CDAM `deleteDocuments` inside that boundary. Moving document deletion out of `PcsCaseService.deleteCase` helps, but HTTP work can still hold a DB transaction open here. Would it be worth keeping mark/confirm and CDAM deletes outside the transaction, and only wrapping the local DB deletes?

src/main/resources/application.yaml
andrewdami11: I still see `discard-after-days: -1` and `FIXED_DELAY|60s` in `application.yaml` with no AAT/prod chart override. Which commit addresses this?

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/CcdCaseDataDeletionServiceTest.java
```diff
@@ -0,0 +1,240 @@
...
+                .thenThrow(feignException);
+        when(objectMapper.valueToTree(any(PCSCase.class))).thenReturn(new ObjectMapper().readTree("{}"));
+
+        // When & Then
+        assertThrows(FeignException.class, () -> underTest.markCaseForDeletion(caseRef));
+
+    }
+
+    @Test
+    void shouldThrpwCcdCaseNotFoundExceptionIfCaseIDFoundToBeInvalid() throws Exception {
```
tvr-hmcts: Minor typo

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/CcdCaseRepository.java
andrewdami11: Looking at the CTE though, `case_event_significant_items` is filtered with `WHERE case_event_id IN (SELECT case_event_id FROM deleted_case_event_audit)`. 
That only includes event IDs that **had** an audit row. If an event has significant items but no audit row, those rows would not be deleted. Would it be worth keying that final delete off deleted_case_event IDs instead?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/CcdCaseRepository.java
```diff
@@ -0,0 +1,67 @@
...
+                   FROM ccd.case_data cd
+                   WHERE cd.created_date < now()::date - :discardDaysAfter
+                   AND cd.state in ('AWAITING_SUBMISSION_TO_HMCTS', 'PENDING_CASE_ISSUED')
+                """,
+                namedParameters,
+                (rs, rowNum) -> rs.getLong("reference")
+        );
+    }
+
+    public List<Long> findExpiredDraftCasesInDraftDiscardedState() {
```
andrewdami11, line 37: This query picks up every case already in `DRAFT_DISCARDED`, with no age filter and no batch size. That is useful for retry cleanup, but a big set could make each sweep heavy. Would it be worth only fetching a limited batch each time?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,91 @@
...
+
+    protected void runSweep() {
+        MDC.put(MDC_TASK_NAME, CASE_DELETION_TASK_NAME);
+        try {
+            List<Long> caseReferences = ccdCaseDataDeletionService.findExpiredDraftCases(discardAfterDays).stream()
+                    .toList();
+            if (!caseReferences.isEmpty()) {
+                log.debug("Found {} expired draft cases to delete", caseReferences.size());
+
+                List<CompletableFuture<Void>> futures = caseReferences.stream()
```
andrewdami11: This starts one background task per expired case, with no limit on how many can run at the same time. A large backlog could put a lot of load on the app, or still be running when the next scheduled sweep starts. Would it be worth capping how many run at once, or only fetching a batch of cases each time?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,91 @@
...
+                        .map(caseRef -> CompletableFuture.runAsync(() ->
+                                        caseDeletionService.performCaseDeletionTasks(caseRef))
+                                .orTimeout(60, TimeUnit.SECONDS)
+                                .exceptionally(ex -> {
+                                    log.error("Case deletion timed out or failed for case: {}", caseRef, ex);
```
tvr-hmcts: Need to apply the taskname and other indication to this log line and any within caseDeletionService.performCaseDeletionTasks(...) as it is a seperate thread so they will not output the same.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,91 @@
...
+    protected void runSweep() {
+        MDC.put(MDC_TASK_NAME, CASE_DELETION_TASK_NAME);
+        try {
+            List<Long> caseReferences = ccdCaseDataDeletionService.findExpiredDraftCases(discardAfterDays).stream()
+                    .toList();
+            if (!caseReferences.isEmpty()) {
+                log.debug("Found {} expired draft cases to delete", caseReferences.size());
+
+                List<CompletableFuture<Void>> futures = caseReferences.stream()
+                        .map(caseRef -> CompletableFuture.runAsync(() ->
```
tvr-hmcts: It will be better to have a dedicated Executor to run these through rather than the default forkjoin which it will use the moment I think if you dig into the runAsync() method.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,91 @@
...
+                                        caseDeletionService.performCaseDeletionTasks(caseRef))
+                                .orTimeout(60, TimeUnit.SECONDS)
```
tvr-hmcts: If you look inside this ... the thread continues.   It will zombie the thread I think which will keep hold of the resources.  If enough threads hand / slow then the impact is application wide.

If you have a seperate ExecutorService (mentioned above) then you can set the time out on each Future callable and let them destroy themselves.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,91 @@
...
+    @Bean
+    public RecurringTask<Void> caseDeletionTask() {
+        return Tasks.recurring(CASE_DELETION_TASK_NAME, Schedules.parseSchedule(schedule))
+                .execute((taskInstance, executionContext) -> runSweep());
+    }
+
+    protected void runSweep() {
+        MDC.put(MDC_TASK_NAME, CASE_DELETION_TASK_NAME);
+        try {
+            List<Long> caseReferences = ccdCaseDataDeletionService.findExpiredDraftCases(discardAfterDays).stream()
```
tvr-hmcts: Worries me on the amount this maybe coming from here.  Perhaps run in batches and just schedule the task to run a little more frequently.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/casedeletion/CcdCaseDataDeletionService.java
```diff
@@ -0,0 +1,98 @@
...
+                    String.valueOf(caseRef),
+                    eventId.name());
+
+            CaseDataContent submitContent = CaseDataContent.builder()
+                    .event(Event.builder().id(eventId.name()).build())
+                    .eventToken(startEventResponse.getToken())
+                    .data(toJsonNode(PCSCase.builder().build()))
+                    .build();
+
+            CaseResource caseResource = coreCaseDataApi.createEvent(idamToken, serviceAuthorization,
```
tvr-hmcts, line 83: I recall this pattern from the MakeAClaim callback.   Praveen said it was awkward and hoped there would be an alternative.   However it is in the back burner.   Just mentioning here as it reminded me.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,91 @@
...
+                    .toList();
+            if (!caseReferences.isEmpty()) {
+                log.debug("Found {} expired draft cases to delete", caseReferences.size());
+
+                List<CompletableFuture<Void>> futures = caseReferences.stream()
+                        .map(caseRef -> CompletableFuture.runAsync(() ->
+                                        caseDeletionService.performCaseDeletionTasks(caseRef))
```
tvr-hmcts: These need to be killable.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,91 @@
...
+@Component
+public class CaseDeletionScheduledTask {
+
+    private static final String CASE_DELETION_TASK_NAME = "case-deletion-task";
+    private static final String MDC_TASK_NAME = "taskName";
+
+    private final String schedule;
+    private final int discardAfterDays;
+    private final CcdCaseDataDeletionService ccdCaseDataDeletionService;
+    private final CaseDeletionService caseDeletionService;
```
tvr-hmcts, line 38: Have a dedicated ExecutorService at this level.

tvr-hmcts (commented):
Some thoughts.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/casedeletion/CaseDeletionService.java
guygrewal77: The document deletion relies on the PcsCaseEntity so it needs to be done before the PcsCase is deleted itself.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: Being addressed, currently testing the new changes.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: Will be setting a batchLimitSize for the queries. Yep it can be run more frequently.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: Being addressed in new changes.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/CcdCaseRepository.java
guygrewal77: Being addressed in new changes.

src/main/resources/application.yaml
guygrewal77: Might need to leave this until the changes are approved as the local changes can easily be committed while testing.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: Updated in latest.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: Fixed in latest.

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/CcdCaseDataDeletionServiceTest.java
guygrewal77: Fixed

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: Fixed.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: Fixed.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: Fixed - will run in small batches until all the cases get processed from the Sql query.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: Fixed.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/CcdCaseRepository.java
guygrewal77: The processing of the cases will be done in small batches to not overwhelm the resources available.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/CcdCaseRepository.java
guygrewal77: Fixed in latest.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
```diff
@@ -97,6 +99,23 @@ public PcsCaseEntity loadCase(long caseReference) {
...
+    public void deleteCase(long caseReference) {
+        PcsCaseEntity pcsCaseEntity = loadCase(caseReference);
+        pcsCaseRepository.delete(pcsCaseEntity);
+        log.debug("Deleted case with reference: {}", caseReference);
+    }
+
+    @Transactional
```
andrewdami11, line 115: `deleteDocuments` is now `@Transactional`, and it still calls CDAM over HTTP inside that method. Moving document deletion out of `deleteCase` helped, but the HTTP calls are back inside a DB transaction here. Would it be worth loading the document URLs first, then deleting from CDAM outside the transaction?

src/main/resources/application.yaml
```diff
@@ -269,6 +269,15 @@ bulk-print:
   # BULK_PRINT_LOOKBACK_HOURS (lower envs only): sweep ignores packs whose documents are older than N hours,
   # so enabling the flag cannot flood send-letter with the historical backlog. Unset in prod = full backlog.
 
+expired-case-deletion:
+  schedule: ${EXPIRED_CASE_DELETION_SCHEDULE:DAILY|03:00}
+  discard-after-days: ${EXPIRED_CASE_DELETION_DISCARD_AFTER_DAYS:30}
+  core-pool-size: ${EXPIRED_CASE_DELETION_CORE_POOL_SIZE:2}
+  max-pool-size: ${EXPIRED_CASE_DELETION_MAX_POOL_SIZE:2}
+  batch-size: ${EXPIRED_CASE_DELETION_BATCH_SIZE:2}
+  queue-capacity: ${EXPIRED_CASE_DELETION_QUEUE_CAPACITY:0}
```
andrewdami11: `queue-capacity` defaults to `0` while `batch-size` and `max-pool-size` default to `2`. With a zero queue, submitting more work than the pool can run at once can reject tasks. Would it be worth setting a small queue by default, or documenting that `batch-size` must stay at or below `max-pool-size` when the queue is `0`?

charts/pcs-api/values.preview.template.yaml
```diff
@@ -29,12 +29,12 @@ java:
     PCS_DB_HOST: "{{ .Release.Name }}-postgresql"
     # HDPI-8115: preview shares one small flexible server across ~70 releases;
     # shrink per-pod pools so idle stacks hold no connections.
-    PCS_DB_POOL_MIN_IDLE: "0"
-    PCS_DB_POOL_MAX_SIZE: "2"
+    PCS_DB_POOL_MIN_IDLE: "4"
+    PCS_DB_POOL_MAX_SIZE: "20"
```
tvr-hmcts, line 33: This would need to be a percentage of the hikari pool size as that is the actual boundary will be within this work load.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,141 @@
...
+        this.schedule = schedule;
+        this.discardAfterDays = discardAfterDays;
+        this.ccdCaseDataDeletionService = ccdCaseDataDeletionService;
+        this.caseDeletionService = caseDeletionService;
+        this.taskTimeoutSeconds = taskTimeoutSeconds;
+        this.batchSize = batchSize;
+
+        ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+        executor.setCorePoolSize(corePoolSize);
+        executor.setMaxPoolSize(maxPoolSize);
```
tvr-hmcts: This would need to be a % of db pool size.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,141 @@
...
+            log.debug("Completed cleanup sweep successfully.");
+        } finally {
+            MDC.remove(MDC_TASK_NAME);
+        }
+    }
+
+    private void processBatch(List<Long> batch, LongConsumer caseAction) {
+        List<CompletableFuture<Void>> futures = batch.stream()
+                .map(caseRef -> CompletableFuture.runAsync(() -> caseAction.accept(caseRef), deletionExecutor)
+                        .orTimeout(taskTimeoutSeconds, TimeUnit.SECONDS)
```
tvr-hmcts: This will not be friendly - it gives you the appearance of being nice ... but things still happen in the background.   It just allows the join() to be happy.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,141 @@
...
+                                     CcdCaseDataDeletionService ccdCaseDataDeletionService,
+                                     CaseDeletionService caseDeletionService) {
+        this.schedule = schedule;
+        this.discardAfterDays = discardAfterDays;
+        this.ccdCaseDataDeletionService = ccdCaseDataDeletionService;
+        this.caseDeletionService = caseDeletionService;
+        this.taskTimeoutSeconds = taskTimeoutSeconds;
+        this.batchSize = batchSize;
+
+        ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
```
tvr-hmcts: There is more happening than on first inspection.   This will be limited by the Hiraki pool.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/casedeletion/CaseDeletionService.java
```diff
@@ -0,0 +1,34 @@
...
+@AllArgsConstructor
+@Slf4j
+public class CaseDeletionService {
+
+    private final CcdCaseDataDeletionService ccdCaseDataDeletionService;
+    private final DraftCaseDataService draftCaseDataService;
+    private final PcsCaseService pcsCaseService;
+
+    public void deleteDocuments(long caseReference) {
+        pcsCaseService.deleteDocuments(caseReference);
```
tvr-hmcts: Need this hooked into the earlier flow with virtuals.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
tvr-hmcts: Though I would add to this in a moment as we can adjust things around a bit.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,141 @@
...
+    public void performCaseDeletionTasks(long caseRef) {
+        log.debug("Performing case deletion tasks for case: {}", caseRef);
+        try {
+            ccdCaseDataDeletionService.markCaseForDeletion(caseRef);
+            ccdCaseDataDeletionService.confirmCaseDisposal(caseRef);
+        } catch (CcdCaseNotFoundException e) {
+            log.error("Case not found in main ccd datastore. Will proceed to delete in decentralised ccd schema");
+        } catch (FeignException e) {
+            log.error("Error occurred while performing case deletion tasks for case: {}. Error: {}",
+                    caseRef, e.getMessage());
```
tvr-hmcts: Pls do not put e.getMessage() in log output - these will be caught later once I finally get my PR over the line.  You can just put the exception at the end and that will be safer through the redaction work.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,141 @@
...
+        this.deletionExecutor = executor;
+    }
+
+    @Bean
+    public RecurringTask<Void> caseDeletionTask() {
+        return Tasks.recurring(CASE_DELETION_TASK_NAME, Schedules.parseSchedule(schedule))
+                .execute((taskInstance, executionContext) -> runSweep());
+    }
+
+    protected void runSweep() {
```
tvr-hmcts: In this method there are two things happening.  I would seperate those.  The first spin I think should use virtual threads as it is mainly jumping then network.   With that we need some constraints as we are only using the default configuration of the HTTP client pool at the moment also.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,141 @@
...
+        try {
+            List<Long> expiredDraftCases = ccdCaseDataDeletionService.findExpiredDraftCases(discardAfterDays);
+            if (!CollectionUtils.isEmpty(expiredDraftCases)) {
+                log.info("Processing {} expired draft cases", expiredDraftCases.size());
+                Lists.partition(expiredDraftCases, batchSize)
+                        .forEach(batch -> processBatch(batch, this::performCaseDeletionTasks));
+            }
+            log.debug("Completed case deletion sweep successfully.");
+
+            List<Long> discardedDraftCases = ccdCaseDataDeletionService.findExpiredDraftCasesInDraftDiscardedState();
```
tvr-hmcts: Is there an actual a hard dependency between this and the processBatch(..., performCaseDeletionTasks)?  I do not think there is so I would take these out of the threading completely and just iterate over them.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,141 @@
...
+            log.error("Error occurred while performing case deletion tasks for case: {}. Error: {}",
+                    caseRef, e.getMessage());
+            throw e;
+        }
+        cleanupDiscardedDraftCases(caseRef);
+    }
+
+    private void cleanupDiscardedDraftCases(long caseRef) {
+        caseDeletionService.deleteDocuments(caseRef);
+        caseDeletionService.deleteCase(caseRef);
```
tvr-hmcts: Can this deleteCase be seperated from deleteDocuments pls.   Then call the deleteDocuments in the virtual threads of the performCaseDeletionTasks() and keep the database work in the iteration after the virtuals have completed.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,141 @@
...
+    private Runnable wrapWithMdc(Runnable task) {
+        var contextMap = MDC.getCopyOfContextMap();
+        return () -> {
+            if (contextMap != null) {
+                MDC.setContextMap(contextMap);
+            }
+            try {
+                task.run();
+            } finally {
+                MDC.clear();
```
tvr-hmcts: Minor ish: The wrapping is cool but can we be sure on if the clear here is good or if it should restore back to the original.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/casedeletion/CaseDeletionService.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    @Transactional(timeout = 15)
+    public void deleteCase(long caseReference) {
```
tvr-hmcts: Problems can happen here in the threading space.  The ON DELETE CASCADE will cause locks in the DB.  Are there shared foreign keys in the tables at all?   This is one of the reasons I think the DB work should iterate instead (along with the db pool sizing)

tvr-hmcts (commented):
Some thoughts.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/CcdCaseRepository.java
andrewdami11: Thanks, looks good.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/casedeletion/CaseDeletionService.java
guygrewal77: This job will be run out of hours in the early hours so won't be affected by any users logging in.

As long as we can configure the settings so that performance is optimal and no issues arise from any performance testing.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: When deleting documents, we need the PcsCaseEntity to retrieve the documents so it needs to run before the deleteCase().

And if it fails, then deleteCase shouldn't be run.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/PcsCaseService.java
andrewdami11: Thanks — the split into `getDocuments` and `deleteDocumentsFromCdam` looks good.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+
+    private void cleanupDiscardedDraftCases(long caseRef) {
+        try {
+            caseDeletionService.deleteDocuments(caseRef);
+            caseDeletionService.deletePcsCase(caseRef);
+            caseDeletionService.deleteCcdCase(caseRef);
+            caseDeletionService.deleteDraftData(caseRef);
+        } catch (DocumentDeletionException e) {
```
andrewdami11: If one of these deletes fails, the code only logs the error and then stops. The later deletes for that case never run. Next time the job runs, the same early step can fail again (for example because the PCS case is already gone), so the leftover CCD or draft data may never get cleared. Should we be carrying on with the later deletes after a step fails, or treat "already deleted" as fine?

charts/pcs-api/values.preview.template.yaml
guygrewal77: What should the hikari pool size be ? These need to be discussed.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: There's been some refactoring so that the deletions from the different schemas are all tied to a single transaction to allow rollbacks if any one fails.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: Refactored so that we're not relying on the entity for the document deletion from cdam. But the document deletion needs to be done just before the case data deletion so we don't end up with the documents deleted but the case data still there.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: How do you mean ? The processBatch is used for two tasks.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: The documents need to be deleted just before the case data deletion. Having a timeout will allow the individual case task to finish in a predictable time with or without error.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: Fixed.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: What would the values need to be ?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
guygrewal77: What if processed sequentially by having the batchSize as 1 ?

src/main/resources/application.yaml
guygrewal77: These values need to be discussed.

andrewdami11: Now that the latest logic/commit has moved to _virtual threads_ and _semaphores_, `queue-capacity` / `core-pool-size` / `max-pool-size` I think are now unused? 
If that's the case, can we drop them?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
```diff
@@ -0,0 +1,168 @@
...
+        return new ArrayList<>(failed);
+    }
+
+    private void completeCaseDeletion(long caseRef) throws InterruptedException {
+        log.debug("Performing case deletion tasks for case: {}", caseRef);
+        ccdCallThrottle.acquire();
+        try {
+            ccdCaseDataDeletionService.markCaseForDeletion(caseRef);
+            ccdCaseDataDeletionService.confirmCaseDisposal(caseRef);
+        } catch (CaseNotFoundException e) {
```
andrewdami11: This catch looks for `CaseNotFoundException`, but the mark/confirm events throw `CcdCaseNotFoundException`, which is a different type. If the case is already gone in the central CCD store, this catch will not run, and the local PCS/draft/CCD cleanup below will not happen. An earlier version of this method caught `CcdCaseNotFoundException` on purpose so cleanup could still continue. Would it be worth catching that type again?

tvr-hmcts: Nice catch (pun intended).

src/main/resources/application.yaml
tvr-hmcts: Yer they can be dropped.   Ive defaulted the Semaphore throttles in the code but will they can be applied instead.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/task/CaseDeletionScheduledTask.java
tvr-hmcts: Updated.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/repository/CcdCaseRepository.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public List<Long> findExpiredDraftCases(int discardAfterDays, int limit) {
+        SqlParameterSource namedParameters = new MapSqlParameterSource()
+            .addValue("discardDaysAfter", discardAfterDays)
+            .addValue("limit", limit);
+        return namedParameterJdbcTemplate.query(
+            """
+                SELECT cd.reference
+                FROM ccd.case_data cd
+                WHERE cd.created_date < now()::date - :discardDaysAfter
```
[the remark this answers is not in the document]
guygrewal77: Fixed in latest. Will be using last_state_modified_date instead.

andrewdami11 (approved):

charts/pcs-api/Chart.yaml
```diff
@@ -51,3 +51,7 @@ dependencies:
     version: ~1.1.0
     repository: 'oci://hmctsprod.azurecr.io/helm'
     condition: wa.enabled
+  - name: ccd-case-disposer
```
arun-hmcts: I think we should remove this entry before we merge to AAT. normal PR doesn't need case disposer ..

tvr-hmcts:
NB: I have been in this and altered a small part of it so I technically should not approve.   I approve the rest.

tvr-hmcts (approved):
Approving in order to move this forward now for DA/QA visibility.

PR #2281, opened 2026-07-28
HDPI-6593: WA task to review gen app additional docs

scottstewart-sl:
### Jira link

See [HDPI-6593](https://tools.hmcts.net/jira/browse/HDPI-6593)

### Change description

Add a new Work Allocation task which is created when a user uploads some additional documents that are associated with a general application.

This PR also introduces dynamic descriptions on the tasks, so that we can put specific data in there rather than having a fixed message. In this case, it is used to add the document names that need reviewing, along with some basic details about the the related gen app and party

image

Finally, I've added a docker compose and README instructions for running WA locally, (for development purposes)

Here is a screenshot of the changed area in the configuration DMN:

image

### Testing done

Unit tests updated and extended. End to end test carried out in preview from CUI.

  * [ ] Yes
  * [x] No

### Checklist


docker-compose-wa.yml
```diff
@@ -0,0 +1,36 @@
+services:
+  camunda-local:
+    platform: linux/amd64
+    image: hmctssandbox.azurecr.io/camunda-local:latest
+    container_name: camunda-local
+    ports:
+      - "8097:8097"
+  wa-workflow-api:
```
gmmagruder, line 8: Think the wa-workflow-api image is not needed since the only other services that are interacted with are wa-task-monitor, wa-task-mangement-api and camunda

scottstewart-sl: I think it's used for the idempotency check in the bpm:

image

Without it the task processes just stop at that stage. I've checked the logs of that container on my machine and I can see it processing the tasks as they go through:

```
2026-07-30T13:26:50.949 INFO  [TopicSubscriptionManager] u.g.h.r.w.c.s.i.IdempotencyTaskWorkerHandler No idempotencyKey found for process instance(4f1572f2-8c1a-11f1-95e4-0242ac130003), probably a service other than wa/ia is using the BPM.
2026-07-30T13:29:14.624 INFO  [TopicSubscriptionManager] u.g.h.r.w.c.s.i.IdempotencyTaskWorkerHandler checking idempotency...
2026-07-30T13:29:14.627 INFO  [TopicSubscriptionManager] u.g.h.r.w.c.s.i.IdempotencyTaskWorkerHandler idempotentId is blank
2026-07-30T13:29:14.628 INFO  [TopicSubscriptionManager] u.g.h.r.w.c.s.i.IdempotencyTaskWorkerHandler No idempotencyKey found for process instance(a40c3e87-8c1a-11f1-95e4-0242ac130003), probably a service other than wa/ia is using the BPM.
```

gmmagruder (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/workallocation/TaskDescriptionService.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Map;
+
+@Service
+@RequiredArgsConstructor
+public class TaskDescriptionService {
+
+    private final PartyService partyService;
+    private final PebbleEngine pebbleEngine;
+
+    public String createGenAppAdditionalDocumentsDescription(long caseReference,
```
benouaer, line 27: given the generic way you've implemented this, using the pebble template, I'm guess you probably won't need a separate method each time you want to customise the task description 
in which case, this method name might be made more generic

src/main/resources/dmn/wa-task-configuration-possession_service.dmn
```diff
@@ -1,12 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
benouaer, line 1: not sure if it might be worth adding a screenshot of the DMN to make it easier to review

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/workallocation/TaskDescriptionService.java
scottstewart-sl: Yes, I'm expecting this class to evolve as we add more task types and their descriptions. For example, the three gen app tasks can probably share a single method and probably a single template. But I will refactor that when we work on those tasks.

src/main/resources/dmn/wa-task-configuration-possession_service.dmn
```diff
@@ -364,265 +429,33 @@ Short name</description>
...
         <inputEntry id="UnaryTests_1kieu3i">
-          <text>"ReviewAdditionalDocumentsGenApp"</text>
+          <text>-</text>
         </inputEntry>
         <outputEntry id="LiteralExpression_1u9r7ki">
           <text>"description"</text>
         </outputEntry>
         <outputEntry id="LiteralExpression_0r3qzf3">
-          <text>"Review the additional documents submitted for the general application, check whether any further case action is required, and take the appropriate action. Only mark the task as complete once the documents have been reviewed and any required action has been completed."</text>
+          <text>taskAttributes.taskDescription</text>
```
benouaer, line 446: are these two different ways of setting the task description, either through description or taskAttributes.description?

src/main/resources/dmn/wa-task-configuration-possession_service.dmn
```diff
@@ -674,6 +513,9 @@ Short name</description>
         <inputEntry id="UnaryTests_0mn0gpa">
           <text></text>
         </inputEntry>
+        <inputEntry id="UnaryTests_1jvkqso">
```
benouaer, line 516: are these empty because we're just doing two tasks?

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentServiceTest.java
```diff
@@ -63,20 +67,32 @@
...
+
     @Captor
     private ArgumentCaptor<List<DocumentEntity>> documentEntityListCaptor;
 
     private DocumentService underTest;
 
     @BeforeEach
     void setUp() {
-        underTest = new DocumentService(documentRepository, documentIdExtractor, documentNameService);
+        when(pcsCase.getCaseReference()).thenReturn(CASE_REFERENCE);
```
benouaer, line 94: should pcsCase be populated instead of being mocked

src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/PcsCaseEntity.java
```diff
@@ -114,6 +114,10 @@ public class PcsCaseEntity {
     @Builder.Default
     private List<CaseFlagEntity> caseFlags = new ArrayList<>();
 
+    public ClaimEntity getMainClaim() {
```
tvr-hmcts, line 122: Minor: I can not see this used.  Also would it be an Optional<ClaimEntity> return type?

benouaer (approved):

src/main/resources/dmn/wa-task-configuration-possession_service.dmn
scottstewart-sl: Good idea - I've attached a screenshot to the PR description now

src/main/resources/dmn/wa-task-configuration-possession_service.dmn
scottstewart-sl: Yes, the output values can basically be created from:

- Fixed content
- CCD case data
- Attributes from the initial request we make to Camunda to create the Camunda task, which are basically [these](https://github.com/hmcts/pcs-api/blob/HDPI-6593_gen_app_additional_docs_wa_task/src/main/java/uk/gov/hmcts/reform/pcs/camunda/CamundaService.java#L96-L111), but note that a few of them get stripped out before they get here.

By using values from the request, we can configure them in Java and make them dynamic if required.

tvr-hmcts (approved):

src/main/resources/dmn/wa-task-configuration-possession_service.dmn
scottstewart-sl: These were added by Camunda Modeller when I added a new input column for the task attributes. (I noticed that the WA code passes the CCD data and the request task attributes when evaluating this DMN). These are empty because we don't want any conditions to apply on the rule, but I _think_ we need the input to be declared so we can reference it in the output, (like we do for the CCD data).

Image

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/document/DocumentServiceTest.java
scottstewart-sl: There's probably not much difference between the approaches. I generally go with a mock when I want to change the return values in different tests, or if I want to verify something was set. In this case, the pcsCase was already a mock, but just a local variable in each test, so I pulled it out to the field level to reduce duplication.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/entity/PcsCaseEntity.java
scottstewart-sl: Good point - I refactored the code to pass the main claim in from the calling class instead of getting it from the case, but forgot to remove this. I've removed it now.

PR #2033, opened 2026-06-26
HDPI-7213: Add CftLibTest for citizen gen app creation

scottstewart-sl:
### Jira link

See [HDPI-7213](https://tools.hmcts.net/jira/browse/HDPI-7213)

### Change description

Add a CftLibTest for citizen gen app creation. This includes some refactoring of the Cftlib code to have more re-usable setup methods, and some shared wiremock stubs for the document services. The Citizen gen app test also runs up its own wiremock so that email notifications can be checked. The main test class that has been added is `CitizenMakeAnApplicationTest`.

The only change to production code is to make the Notify API URL come from config rather than using the default one from the client library.

There are new helper services/methods for the CftLib tests to:

* Create a minimal case
* Pay for a claim, (by simulating the payment API callback)
* Wait for a case to change state
* Get defendant access codes
* Link a citizen user to a case, with an access code
* Check what notifications have been sent

### Testing done

Build run.

### Security Vulnerability Assessment ###

**CVE Suppression:** Are there any CVEs present in the codebase (either newly introduced or pre-existing) that are being intentionally suppressed or ignored by this commit?
  * [ ] Yes
  * [x] No

### Checklist

- [x] commit messages are meaningful and follow good commit message guidelines
- [ ] README and other documentation has been updated / added (if needed)
- [x] tests have been updated / new tests has been added (if needed)
- [ ] Does this PR introduce a breaking change

PR #2303, opened 2026-07-30
HDPI-5840 - Notice of Change - Informing of representation - Defendant LiP

mazzopardi2:
### Jira link

See [HDPI-5840](https://tools.hmcts.net/jira/browse/HDPI-5840)

### Change description
- New email template and notification claim type.
- NotificationService.sendNoticeOfChangeCompletedEmailNotification, triggered from LegalRepresentativePartyLinkService once the legal representative has been linked to the party.
- Adds NOTIFY_URL to the preview chart values

Screenshot 2026-08-03 at 08 23 29

### Testing done
- Unit tests added and existing tests extended.
- Manually verified end to end on the PR preview environment: created a case, submitted a defendant response with an email address, then completed NoC via XUI.

  * [ ] Yes
  * [x] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationPersonalisationFactory.java
```diff
@@ -69,6 +75,28 @@ public CounterclaimPaymentSuccessPersonalisation counterclaimSuccess(DefendantRe
             .build();
     }
 
+    public NoticeOfChangeCompletedPersonalisation noticeOfChangeCompleted(PartyEntity partyEntity,
```
guygrewal77, line 78: Could you have a look at adding some test coverage for this method ?

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
```diff
@@ -155,13 +155,23 @@ public void sendGenAppReceivedEmail(GenAppEntity genAppEntity) {
         );
     }
 
+    public EmailNotificationResponse sendNoticeOfChangeCompletedEmailNotification(PartyEntity defendant) {
```
guygrewal77: Can you look into adding some test coverage here please ?

src/main/java/uk/gov/hmcts/reform/pcs/notify/template/personalisation/NoticeOfChangeCompletedPersonalisation.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+@Builder
+public class NoticeOfChangeCompletedPersonalisation implements TemplatePersonalisation {
+    private final BasePersonalisation base;
+    private final String address;
+
+    @Override
```
guygrewal77, line 13: Can we have some test coverage for this class please ?

tvr-hmcts (commented):
Is looking ok.   Agree with Guy's comments on some additional test coverage - and one fails for me at the moment - 

@DisplayName("Should include base fields and the formatted property address")

in NotificationPersonalisationFactoryTest

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
tvr-hmcts: Does the EmailNotificationResponse need to be returned?

charts/pcs-api/values.preview.template.yaml
```diff
@@ -48,6 +48,7 @@ java:
     SEND_LETTER_URL: http://rpe-send-letter-service-aat.service.core-compute-aat.internal
     IDAM_API_URL: https://idam-api.aat.platform.hmcts.net
     IDAM_S2S_AUTH_URL: http://rpe-service-auth-provider-aat.service.core-compute-aat.internal
+    NOTIFY_URL: https://api.notifications.service.gov.uk
```
tvr-hmcts, line 51: Just a question ... was this not needed before?  Wondering why it is applied now.

tvr-hmcts (approved):
LGTM
A question and a minor comment (if I remember correctly)

src/main/java/uk/gov/hmcts/reform/pcs/notify/service/NotificationService.java
mazzopardi2: So nothing consumes it to be honest but all but 1 of the send methods in NotificationService return EmailNotificationResponse, so I followed the existing pattern rather than making this one the exception. I've updated it now though - good point! Is it worth creating a small tidy up ticket for this class?

charts/pcs-api/values.preview.template.yaml
mazzopardi2: Not quite sure - think the API key was wired up which might have made the config look complete? Think this is one of the first few tickets to verify email notifications in preview so perhaps it was overlooked previously as not needed?

jonathanditcher-solirius (approved):

charts/pcs-api/values.preview.template.yaml
scottstewart-sl: This was missing from values.yaml after a recent change I made here:
https://github.com/hmcts/pcs-api/pull/2033

I'll add it into values.yaml after testing it, and then you won't need this override in the `preview` values, although it won't cause any harm.

PR #2308, opened 2026-07-31
Hdpi 7379 save defendant case flags

toby-plunkett:
### Jira link

See [HDPI-7379](https://tools.hmcts.net/jira/browse/HDPI-7379)

### Change description
This PR adds our integration with CUI Your Support - Reasonable Adjustments. 

**In Summary:** 
Frontend: [here](https://github.com/hmcts/pcs-frontend/pull/1757/)
- Allows us to navigate to Your Support Microsite via Check your answers on Respond to claim. 
- Retrieves the Adjustments requested (as flags)
- Posts them to Draft Database after submission & at end of Respond to claim journey. 
 
Backend: [here](https://github.com/hmcts/pcs-api/pull/2284)
- Allows Reasonable adjustments to be saved and sent as draft data. 
- Persists Reasonable adjustments as case flags. 
- Adds Unique constraint to flag_ref_data to limit potential for exploitation. 

### Testing done
Many unit tests have been added. 
As well as manual testing. A video will be also provided.

  * [ ] Yes
  * [x] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -110,10 +157,19 @@ private <T extends BaseCaseFlag> List<T>  mergeFlagDetails(Flags incomingCaseFla
...
-                                               String visibility) {
+                                               String visibility,
+                                               RefDataPolicy refDataPolicy) {
+
+        Optional<FlagRefDataEntity> existingFlagRefData =
+            flagRefDataRepository.findByFlagCode(incomingFlagDetail.getFlagCode());
 
-        FlagRefDataEntity flagRefDataEntity = flagRefDataRepository.findByFlagCode(
-            incomingFlagDetail.getFlagCode()).orElse(new FlagRefDataEntity());
+        // Reference data is shared by every case using this flag code, so a party-supplied flag
```
toby-plunkett, line 195: inline with the new migration comment seen later in pr

src/main/resources/db/migration/V012__add_unique_flag_code_constraint.sql
```diff
@@ -0,0 +1,44 @@
+-- flag_ref_data holds one row per flag code, shared by every case using that flag. Nothing stopped
```
toby-plunkett, line 1: This change enforces a unique constraint on our ref data flags. This means, if a flag came in with some spoofed data, it would not be updated as the previous (correct) flag would block it.

src/integrationTest/java/uk/gov/hmcts/reform/pcs/ccd/repository/FlagRefDataCodeConstraintIT.java
```diff
@@ -0,0 +1,51 @@
...
+@SpringBootTest
+@ActiveProfiles("integration")
+@DisplayName("flag_ref_data - one row per flag code")
+class FlagRefDataCodeConstraintIT extends AbstractPostgresContainerIT {
+
+    @Autowired
+    private FlagRefDataRepository flagRefDataRepository;
+
+    @Test
+    @DisplayName("rejects a second row for a flag code already held")
```
gergelykisshmcts, line 26: The @Container is static on the shared base, so RA0035/RA0012/PF0015 leak into every other IT in the JVM: with the new unique constraint a collision is now a failure rather than a silent duplicate. @Transactional on the class would roll each test back.

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/ClaimResponseServiceTest.java
```diff
@@ -408,12 +412,124 @@ void shouldLeaveAddressSameAsPropertyUnchangedWhenCorrespondenceAddressConfirmat
...
+        // Given
+        UUID partyId = UUID.randomUUID();
+        Flags defendantFlags = Flags.builder().partyName("Jack Smith").roleOnCase("Defendant").build();
+        PossessionClaimResponse response = buildResponse(
+            Party.builder().build(),
+            DefendantResponses.builder().contactByEmail(VerticalYesNo.YES).build()
+        ).toBuilder()
+            .defendantFlags(defendantFlags)
+            .build();
+
```
gergelykisshmcts, line 503: This body is identical to shouldSaveDefendantFlagsAgainstPartyForCitizen: there's no legal-rep-specific behaviour being tested let's drop it or make the setup actually testing something distinct.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseFlagsView.java
```diff
@@ -88,6 +88,7 @@ private List<ListValue<String>> getPaths(String entityPaths) {
 
         return Arrays.stream(entityPaths.split(PATHS_DELIMITER))
                 .map(pathPairs -> pathPairs.split(PATH_DELIMITER))
+                .filter(paths -> paths.length > 1)
```
gergelykisshmcts, line 92: This silently drops the segment, and "_" and " : "are both legal characters in the path text we now accept from cui-ra so a value like "Note: I need X" round-trips as "Note" if that is a possibility as an input here.

gergelykisshmcts (changes requested):
I did not commented the the caseFlagService as was mentioned in the pr "inline with the new migration comment seen later in pr"
2 comments for that service:
"I no longer need any adjustments" is indistinguishable from "the field wasn't sent", so RA flags can only ever be added, never withdrawn. Is that the intended or does the empty-list case need to reach the removeIf?
CREATE_IF_ABSENT still creates a shared flag_ref_data row from the payload when the code is unseen, so a defendant can insert arbitrary rows with their own name/visibility/availableExternally for anything matching RA*. 
Maybe would worth seed the published RA codes in V004 and skip unknown ones instead, making this policy reference-only.

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/ClaimResponseServiceTest.java
toby-plunkett: good catch, I believe the two paths were merged into a single, but my tests were written to cover both. Will remove this.

addressed - removed.

src/integrationTest/java/uk/gov/hmcts/reform/pcs/ccd/repository/FlagRefDataCodeConstraintIT.java
toby-plunkett: Addressed - Tag added

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseFlagsView.java
toby-plunkett: Addressed - Delimiter upped to '2'. 

Colons are fixed, underscores are not. 
An underscore inside a path value is indistinguishable from the separator, the read side can only choose between crashing and dropping  the fragment.

gergelykisshmcts (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -23,21 +25,64 @@
...
+    }
+
+    /**
+     * Applies the reasonable adjustment flags a defendant supplied via the cui-ra microsite to their
+     * party. Only RA flags are accepted and replaced. Caseworker flags arrive through
+     * {@link #mergePartyFlags(List, Set)} instead, which is not restricted in this way.
+     */
+    public void saveReasonableAdjustmentFlags(PartyEntity partyEntity, Flags incomingFlags) {
+        if (incomingFlags == null || CollectionUtils.isEmpty(incomingFlags.getDetails())) {
+            return;
```
tvr-hmcts, line 67: On the tests that hit this point it would be good to verify that none of the other aspects of the method are triggered.

shouldRetainExistingPartyFlagsWhenNoFlagsSupplied()

tvr-hmcts (approved):
Looks good to me.
One minor to consider.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -110,10 +157,19 @@ private <T extends BaseCaseFlag> List<T>  mergeFlagDetails(Flags incomingCaseFla
...
+                                               RefDataPolicy refDataPolicy) {
+
+        Optional<FlagRefDataEntity> existingFlagRefData =
+            flagRefDataRepository.findByFlagCode(incomingFlagDetail.getFlagCode());
 
-        FlagRefDataEntity flagRefDataEntity = flagRefDataRepository.findByFlagCode(
-            incomingFlagDetail.getFlagCode()).orElse(new FlagRefDataEntity());
+        // Reference data is shared by every case using this flag code, so a party-supplied flag
+        // references the existing row as it stands rather than rewriting it from the payload.
+        if (existingFlagRefData.isPresent() && refDataPolicy == RefDataPolicy.CREATE_IF_ABSENT) {
```
scottstewart-sl, line 197: Just wondering - should this second part of the condition be `refDataPolicy != RefDataPolicy.UPDATE_FROM_PAYLOAD` ? It's functionally the same while there are just 2 values in that enum, (assuming the parameter is never null), but I think it better represents the intent doesn't it?

toby-plunkett: I personally think this reads cleaner e.g x=true && y=true  instead of  x=true && y=!true

But I do see where you're coming from. In the future we will only be reading from the database (and it will be seeded separately every x days), not creating if absent so this semantics will be removed.

PR #2290, opened 2026-07-29
HDPI-7768: retrofit merged event states

sadmanrahh:
### Jira link

See [PROJ-7768](https://tools.hmcts.net/jira/browse/PROJ-7768)

### Change description

The purpose of these changes are to retrofit the states to the previously merged events.

### Testing done

Manual testing local and preview

### Security Vulnerability Assessment ###

**CVE Suppression:** Are there any CVEs present in the codebase (either newly introduced or pre-existing) that are being intentionally suppressed or ignored by this commit?
  * [ ] Yes
  * [X] No

### Checklist

- [X] commit messages are meaningful and follow good commit message guidelines
- [ ] README and other documentation has been updated / added (if needed)
- [X] tests have been updated / new tests has been added (if needed)
- [ ] Does this PR introduce a breaking change

PR #2335, opened 2026-08-04
Enable Case Flags v2.1 for internal users

gergelykisshmcts:
### Jira link
See [HDPI-7765](https://tools.hmcts.net/jira/browse/HDPI-7765)

### Change description

Create and Manage Case Flags journeys for internal users.

It also makes the events available in all relevant post-submission states, fixes the missing access for CTSC and WLU users, and adds the internal/external flag separation needed by the v2.1 component.

This includes the fix for HDPI-7905. The external professional-user journey remains part of HDPI-7368.

Updated the internal Create and Manage Case Flags events to use the v2.1 journeys.
Made both events available from Pending Case Issued onwards, excluding the unsubmitted draft state.
Added the missing state-level access for CTSC and WLU users.
Added separate internal and external party flag collections.
Store visibility against each flag rather than the shared flag reference data.
Updated flag loading and saving so internal and external flags are merged correctly and neither collection can be accidentally removed.
Added the group ID required for ExUI to display both party flag collections together for internal users.

Includes database update.

### Testing done
Unit tests and checkstyle pass.
Integration tests pass.
Generated CCD configuration checked for the new journeys, states, fields and permissions.
Database migration validated against the application entities.

Reviewer and tester notes:
The PR preview needs the pr-values:ccd label so the ExUI screens can be tested.

The existing Playwright Case Flags tests have not yet been updated for the v2.1 screens and may fail until that work is completed.

The automated accessibility scan will also report a known missing-label issue in the shared ExUI Case Flags component. This is not produced by PCS and is being raised with the ExUI team.

  * [ ] Yes
  * [X] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/FlagVisibilityConverter.java
```diff
@@ -0,0 +1,13 @@
+package uk.gov.hmcts.reform.pcs.ccd.util;
+
+import uk.gov.hmcts.ccd.sdk.type.FlagVisibility;
+
+public class FlagVisibilityConverter {
+
+    public static FlagVisibility toFlagVisibility(String visibility) {
```
tvr-hmcts, line 7: Perhaps a test on this.  Recognise it is covered by other tests but worth the enforcement of it with its own.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/Party.java
```diff
@@ -53,4 +53,11 @@ public class Party {
     )
     private Flags defendantFlags;
 
+    @CCD(
+        access = {InternalCaseFlagAccess.class},
```
tvr-hmcts: Is this access correct for 'FlagsExternal'?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/PCSCase.java
```diff
@@ -726,6 +726,12 @@ public class PCSCase {
 
     private FlagLauncher flagLauncherInternal;
 
+    @CCD(
+        access = {InternalCaseFlagAccess.class},
```
tvr-hmcts: Same here?   I put a seperate one on these in 7905 (I could be wrong though obviously)

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/CreateFlags.java
```diff
@@ -44,10 +45,11 @@ public void configureDecentralised(DecentralisedConfigBuilder<PCSCase, State, Us
                 .optional(PCSCase::getParties, ShowConditions.NEVER_SHOW, true, true)
                 .list(PCSCase::getAllDefendants, ShowConditions.NEVER_SHOW)
                     .optional(Party::getDefendantFlags, ShowConditions.NEVER_SHOW, true)
+                    .optional(Party::getDefendantFlagsExternal, ShowConditions.NEVER_SHOW, true)
```
tvr-hmcts, line 59: I was under the impression we would need to add to the .grant(...) above for the event selection to be avaliable for the users mentioned in the 7905.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -52,31 +54,56 @@ public void mergePartyFlags(List<ListValue<Party>> incomingParties, Set<PartyEnt
...
+
+    private void mergePartyFlagCollections(Party incomingParty, PartyEntity partyEntity) {
+        Flags incomingInternalFlags = incomingParty.getDefendantFlags();
+        Flags incomingExternalFlags = incomingParty.getDefendantFlagsExternal();
 
-                partyEntity.getDefendantFlags().clear();
-                partyEntity.getDefendantFlags().addAll(mergedCasePartyFlags);
-            }
+        if (hasNoFlagDetails(incomingInternalFlags) && hasNoFlagDetails(incomingExternalFlags)) {
+            return;
```
tvr-hmcts, line 132: No test hits this exit point.

tvr-hmcts (commented):
Looks good.   Ive made a few comments and will like to test it in preview once the environment has come up (and I guess the playwright has been fixed up)

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/Party.java
benouaer: no, that should be external, so claimant-solicitor/defendant-solicitor access

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/PCSCase.java
benouaer: internal users need access to this as well, but we need external users to have access as well
(unless the ticket is limiting this to internal users at this stage)

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/CaseFlagStates.java
```diff
@@ -0,0 +1,22 @@
+package uk.gov.hmcts.reform.pcs.ccd.event;
+
+import uk.gov.hmcts.reform.pcs.ccd.domain.State;
+
+public final class CaseFlagStates {
```
benouaer, line 5: does this need to be a separate class for just case flags? we can have a generic class with multiple lists of states for case flags and other events

src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/FlagVisibilityConverter.java
gergelykisshmcts: Added FlagVisibilityConverterTest covering External/Internal case-insensitively, plus null, blank and unknown values. The last three all fall back to Internal, which keeps older flags without visibility data working as expected.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/Party.java
gergelykisshmcts: Added ExternalCaseFlagAccess alongside InternalCaseFlagAccess.
Internal access still needs to stay because the internal v2.1 journey can write externally available flags into this collection, and the Case Flags tab reads both collections back together by groupId.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/CreateFlags.java
gergelykisshmcts: The team leader roles do not need adding here. They are covered by RoleToAccessProfiles, which maps them onto the relevant admin profiles at runtime.
The team leader rows in AuthorisationCaseEvent come from grantHistoryOnly, not the event grant itself. So I think no change needed.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/PCSCase.java
gergelykisshmcts: Added ExternalCaseFlagAccess alongside the internal access here as well.
This is a root level case field, so the generated definition now includes CRU for claimant and defendant solicitors. The field is not used by an event or tab yet, so this only becomes active once HDPI-7368 adds the external Support journeys.
Also noticed master already has InternalCaseFlagAccess for CASE_ISSUED from 7905. This branch adds it to the remaining states, so probably worth syncing to make sure we are not overlapping.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
gergelykisshmcts: Good catch. Added two tests covering both cases:
shouldRetainPartyFlagsWhenIncomingPartyHasNoFlagCollections
shouldRetainPartyFlagsWhenIncomingFlagCollectionsAreEmpty
Both verify existing flags are retained and flagRefDataRepository is not touched.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/CaseFlagStates.java
gergelykisshmcts: Fair question, but I’d keep it as it is. CASE_FLAG_STATES is already shared by four Case Flag events, so it represents one clear domain rule rather than a one-off list. Moving it into a generic state class would just group unrelated rules together and make future changes harder to understand.

tvr-hmcts: I think this will end up in another class that is coming in via EventsStates in https://github.com/hmcts/pcs-api/pull/2290

tvr-hmcts (approved):
This looks good to progress to DA.

The e2e for CC label may need applying again and those tests fixing up as there will be differences based on what this does.

toby-plunkett (approved):
lgtm

PR #2336, opened 2026-08-04
Set upload_timestamp on every document surfaced in Case File View

linusnorton:
## What

Case File View 1.1 adds chronological (upload date) and alphabetical sort, both inside folders and in a new flat "date only" view. The sort is driven by the `upload_timestamp` attribute added to the CCD `Document` type in 1.1.

The [CFV 1.1 how-to guide](https://tools.hmcts.net/confluence/spaces/RCCD/pages/1764268327/How+To+Guide+-+Case+File+View+1.1) lists three onboarding steps. This PR covers the PCS side of all three.

## Why CCD can't populate it for us

For a centralised service, `ccd-data-store`'s `CaseDocumentTimestampService` stamps `upload_timestamp` on every event submission: it diffs `document_url` values between the stored case and the modified case, and stamps `now` on anything new.

PCS is decentralised. Our `Document` fields are **read-path projections** — rebuilt from `DocumentEntity` rows every time CCD asks for the case. Anything `ccd-data-store` writes into them is discarded on the next read. So PCS has to set the value itself, which we can do better anyway: `DocumentEntity.submittedDate` is a `@CreationTimestamp` recording the actual upload time, not an inference from a URL diff.

## The gap this fixes

`DocumentsView` already set `uploadTimestamp` on `allDocuments`. The other four views did not.

That matters because `CategoriesAndDocumentsService.buildCategorisedDocumentDictionary()` walks **every** `Document`-typed field on the case, not just `allDocuments`. And `DocumentsView.isNotInCaseDetailsTab()` deliberately *excludes* these types from `allDocuments` (unless they carry a description):

- `TENANCY_AGREEMENT`, `OCCUPATION_LICENCE`
- `POSSESSION_NOTICE`
- `RENT_STATEMENT`
- `ENERGY_PERFORMANCE_CERTIFICATE`, `EICR_REPORT`, `GAS_SAFETY_CERTIFICATE`

So those documents reach Case File View **only** through `ClaimView`, `RentArrearsView`, `NoticeOfPossessionView` and `TenancyLicenceView` — and without this change they'd have arrived with no timestamp and sorted as "older".

## Changes

| File | Change |
|---|---|
| `view/DocumentViewUtil.java` (new) | Shared `Instant` → `LocalDateTime` (UTC) conversion, so the five call sites don't repeat it |
| `view/DocumentsView.java` | Refactored onto the helper — no behaviour change |
| `view/ClaimView.java` | Set `uploadTimestamp` (EPC / gas safety / EICR / Wales docs) |
| `view/RentArrearsView.java` | Set `uploadTimestamp` (`RentArrearsSection.statementDocuments`) |
| `view/NoticeOfPossessionView.java` | Set `uploadTimestamp` (`NoticeServedDetails.documents`) |
| `view/TenancyLicenceView.java` | Set `uploadTimestamp` (tenancy licence + Wales occupation licence docs) |
| `charts/pcs-api/values.ccd.preview.template.yaml` | `UPLOAD_TIMESTAMP_FEATURED_CASE_TYPES: "PCS"` |

## On the whitelist entry

Preview only, as requested. It is **consistency rather than necessity**: the `ccd.upload-timestamp-featured-case-types` flag gates exactly one thing in `ccd-data-store` — the `addUploadTimestamps` call — and for PCS that call's write is discarded regardless. It's here so preview matches the AAT/prod whitelist we'll ask the CCD team to add, and so the config difference doesn't look like an oversight later. `PCS` with no suffix is correct for preview: `CASE_TYPE_SUFFIX` is only set (to `staging`) in `values.aat.template.yaml`.

**Still needed outside this repo:** a request to the CCD team to add PCS to `UPLOAD_TIMESTAMP_FEATURED_CASE_TYPES` in `cnp-flux-config` for AAT and prod, if we want the config to be consistent there too.

## Testing

Unit tests extended in all four view tests to assert the timestamp round-trips as UTC — including both `TenancyLicenceView` paths (England `TENANCY_AGREEMENT` and Wales `OCCUPATION_LICENCE`).

```
./gradlew compileJava compileTestJava checkstyleMain checkstyleTest   BUILD SUCCESSFUL
./gradlew test --tests '*ViewTest'                                    BUILD SUCCESSFUL
```

No backfill of historical documents — matching the product decision taken for CFV 1.1 centrally. Documents uploaded before this ships keep sorting as "older".


src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/DocumentViewUtil.java
```diff
@@ -0,0 +1,24 @@
...
+
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+import uk.gov.hmcts.reform.pcs.ccd.entity.DocumentEntity;
+
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class DocumentViewUtil {
```
tvr-hmcts, line 11: Probably just needs an independent test on this.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/DocumentViewUtil.java
```diff
@@ -0,0 +1,24 @@
...
+    /**
+     * The upload timestamp shown in Case File View. As a decentralised service PCS owns this value,
+     * so it is set on every Document we return to CCD rather than being populated by ccd-data-store.
+     */
+    public static LocalDateTime uploadTimestamp(DocumentEntity documentEntity) {
+        if (documentEntity.getSubmittedDate() == null) {
+            return null;
+        }
+
+        return documentEntity.getSubmittedDate().atZone(ZoneOffset.UTC).toLocalDateTime();
```
tvr-hmcts, line 22: Minor ish thought: The issue here maybe that there are a number of ways that this is set currently.

Within the DocumentEntity itself there is two approaches

I think the @CreationTimestamp is ok

but the @PrePersist setting to Instant.now() perhaps can go.

There is a question mark on the sql 

submitted_date timestamp with time zone DEFAULT CURRENT_TIMESTAMP NOT NULL,

may have a question mark on it being time zone DEFAULT but we are pretty inconsistent with this in the sql - would have been good to align them all when the compression of the sql scripts happened but that was time pressured.

tvr-hmcts (approved):
LGTM
Just a comment on the ways in which this date/time is set at the moment.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/DocumentViewUtil.java
```diff
@@ -0,0 +1,24 @@
...
+import java.time.ZoneOffset;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class DocumentViewUtil {
+
+    /**
+     * The upload timestamp shown in Case File View. As a decentralised service PCS owns this value,
+     * so it is set on every Document we return to CCD rather than being populated by ccd-data-store.
+     */
+    public static LocalDateTime uploadTimestamp(DocumentEntity documentEntity) {
```
scottstewart-sl: Can we have this class as a bean with non-static methods? It makes unit tests a more scoped to the class they are testing, and is generally easier to stub different scenarios in tests.

arun-hmcts (approved):

scottstewart-sl (approved):

PR #2337, opened 2026-08-04
HDPI-8081: delay creation of NEW_CLAIM_CREATE_NEW_HEARING task by one day if gen app expected

gmmagruder:
### Jira link

See [HDPI-8081](https://tools.hmcts.net/jira/browse/HDPI-8081)

### Change description

Delay creation of NEW_CLAIM_CREATE_NEW_HEARING task by 24 hours if gen app is expected
In test environments, is delayed by 5 minutes.

### Testing done

Manual testing on preview and unit testing

  * [ ] Yes
  * [ ] No

### Checklist


src/main/java/uk/gov/hmcts/reform/pcs/camunda/CamundaService.java
```diff
@@ -38,11 +40,23 @@ public class CamundaService {
     private final Clock utcClock;
 
     public void createTask(long caseId, TaskType taskType) {
-        scheduleCamundaRequest(Action.CREATE, caseId, taskType);
+        scheduleCamundaRequest(Action.CREATE, caseId, taskType, Instant.now(utcClock));
+    }
+
+    public void createTask(long caseId, TaskType taskType, Integer daysDelayed) {
```
scottstewart-sl: Can you make the `daysDelayed` parameter into a `Duration` so it is more flexible for other delays if we need them, (e.g.  2 hours), and also to make the meaning of the parameter in the method call more obvious when reading it?

src/main/java/uk/gov/hmcts/reform/pcs/camunda/CamundaService.java
```diff
@@ -38,11 +40,23 @@ public class CamundaService {
...
+        Instant scheduledTo = Instant.now(utcClock);
+
+        if (TestSupportEnvironment.isNonProdTestSupportEnabled()) {
+            scheduledTo = scheduledTo.plus(5, ChronoUnit.MINUTES);
```
scottstewart-sl: I'm not sure about this approach. It would mean that WA tasks that are scheduled for "now" would still get a 5 minute delay on them wouldn't it? You could change it to make it a 5 minute ceiling, but even then, 5 minutes would be impractical for any automated testing. I think a more realistic approach would be to adjust the clock used by the scheduler to make it think that, say, 24 hours had elapsed. 

Can you take a look at https://urgo.medium.com/testing-db-scheduler-in-spring-boot-application-0c9929245bc6 to see if an approach like that could be used? I'm envisaging an endpoint in the TestingSupportController that can be used to shift the (scheduler) clock forward, and re-trigger. Then automated tests, (or manually testers), can advance the time as required.

src/test/java/uk/gov/hmcts/reform/pcs/camunda/CamundaServiceTest.java
```diff
@@ -289,6 +292,50 @@ void shouldSkipCancellingTaskIfWaIsNotEnabled() {
...
+            assertThat(schedulableInstance.getNextExecutionTime(Instant.now()))
+                .isEqualTo(Instant.parse("2025-08-28T12:51:19Z"));
+        }
+    }
+
+    @Test
+    void shouldScheduleCamundaCreateRequestTaskWithDelayPreview() {
+        try (MockedStatic<TestSupportEnvironment> mockedTestSupport = mockStatic(TestSupportEnvironment.class)) {
+            // When
+            mockedTestSupport.when(TestSupportEnvironment::isNonProdTestSupportEnabled).thenReturn(true);
```
sadmanrahh: This test name still says Preview, but the stub now represents any non-prod test-support environment, including AAT / ENABLE_TESTING_SUPPORT. Could we rename it or add parameterized cases so the broader behaviour is explicit?

src/main/java/uk/gov/hmcts/reform/pcs/camunda/CamundaService.java
gmmagruder: Addressed

src/test/java/uk/gov/hmcts/reform/pcs/camunda/CamundaServiceTest.java
gmmagruder: Removed test as will be using api to trigger the scheduled tasks.

src/main/java/uk/gov/hmcts/reform/pcs/camunda/CamundaService.java
gmmagruder: Wasn't able to use settable clock to trigger the db scheduler but was able to add an api endpoint to reschedule tasks based on the original scheduled time.

sadmanrahh (approved):

src/main/java/uk/gov/hmcts/reform/pcs/testingsupport/endpoint/TestingSupportController.java
```diff
@@ -508,4 +511,65 @@ public ResponseEntity<String> setPartyEmail(
...
+        summary = "Reschedule Camunda request",
+        description = "Reschedules Camunda requests in the db scheduler to be triggered immediately"
+    )
+    @ApiResponses(value = {
+        @ApiResponse(responseCode = "200", description = "Camunda request tasks rescheduled successfully"),
+        @ApiResponse(responseCode = "401", description = "Unauthorized - Invalid or missing authorization token"),
+        @ApiResponse(responseCode = "403", description = "Forbidden - Invalid or missing service authorization token"),
+        @ApiResponse(responseCode = "500", description = "Internal server error")
+    })
+    @PostMapping("/reschedule-camunda-request")
```
scottstewart-sl: Minor - this could reschedule more than one request so can you make the path (and summary above) into plural?

src/main/java/uk/gov/hmcts/reform/pcs/testingsupport/endpoint/TestingSupportController.java
```diff
@@ -508,4 +511,65 @@ public ResponseEntity<String> setPartyEmail(
         return ResponseEntity.ok().build();
     }
 
+    @Operation(
+        summary = "Reschedule Camunda request",
+        description = "Reschedules Camunda requests in the db scheduler to be triggered immediately"
```
scottstewart-sl: How about tweaking this to something like "Reschedules all scheduled Camunda requests within the specified window from now to be triggered immediately".

scottstewart-sl (approved):

PR #2340, opened 2026-08-04
Cut preview DB pool sizes for pcs-api and CCD

linusnorton:
## Why

Preview PRs stopped getting their databases created: pods crashlooped on
`FATAL: database "pr-NNNN-pcs" does not exist`. The ASO `FlexibleServersDatabase`
CRs could not reconcile — the GET that precedes the PUT was returning
`502 ServerUnableToRetriveData` for the whole `databases` sub-resource of
`pcs-preview`. 60 of 188 CRs were stuck across ten different PRs.

The server was carrying **1156 connections, 1147 of them idle**, across ~180
databases on a 2 vCPU / 8 GiB `Standard_D2ds_v5`. `max_connections` had been
raised to 5000 (SKU default is 859) to accommodate this. 1099 of those
connections were idle JDBC pool connections.

So the fix is to stop the Java apps holding so many connections open.

## What this changes

**pcs-api** (`values.preview.template.yaml`) — pool capped at 4/1, and
db-scheduler cut to 2 threads on a 60s poll. Each stack was holding 10 idle
connections.

**ccd-data-store-api** (`values.ccd.preview.template.yaml`) —
`DATA_STORE_DB_MAX_POOL_SIZE` 80 → 5, and `DATA_STORE_DB_MIN_IDLE` pinned to 1
(upstream default 8, which is what produced 8 idle connections per stack). The
80 here is the single biggest reason `max_connections` ever needed raising —
it is 5× the upstream default of 16.

**ccd-definition-store-api** — `DEFINITION_STORE_DB_MAX_POOL_SIZE: 2`
(upstream default 5).

## The Hikari config was being silently ignored

`DataSourceConfiguration` built its pool via
`DataSourceProperties.initializeDataSourceBuilder()`, which reads only
`url`/`username`/`password`/`driver-class-name`. The entire
`spring.datasource.hikari` block in `application.yaml` — including
`maximumPoolSize: 10` and `minimumIdle: 2` — never reached the pool.

Confirmed empirically: every stack held exactly 10 idle connections, which is
Hikari's *default* pool size, not the configured 10-max/2-idle.

Without fixing this, the new env vars would have been no-ops. The bean now
builds a `HikariDataSource` with `@ConfigurationProperties("spring.datasource.hikari")`
bound to it. The transaction-aware proxy is kept as a separate `@Primary` bean
so db-scheduler still joins active transactions, and so `SchedulingConfig`'s
by-type `DataSource` injection stays unambiguous.

`db-scheduler.threads` was hardcoded to `10`; it is now overridable, since
each scheduler thread takes a pool connection.

`DataSourceConfigurationTest` covers both halves. Both tests were verified to
fail against the previous code.

## Expected impact

Steady-state for the three services drops from roughly 800 connections to
~385. Worst case per stack goes from 35 × 95 to 35 × 11.

## Not included

WA services are in #2341, since `values.wa.preview.template.yaml` is not on
master yet. ORM and RAS are excluded — they are being removed.

## Note on the stuck CRs

The affected databases were created manually through the data plane, so the
pods recovered, but the CRs still read `Ready=False`. ASO will keep retrying
and will not recreate an existing database, so that is benign — but a
`helm uninstall` on those releases **will** cascade-delete the real database
via the CR finalizer.


charts/pcs-api/values.preview.template.yaml
```diff
@@ -29,6 +29,10 @@ java:
     PCS_DB_HOST: "{{ .Release.Name }}-postgresql"
     PCS_DB_USER_NAME: "{{ .Values.postgresql.auth.username}}"
     PCS_DB_NAME: "{{ .Values.postgresql.auth.database}}"
+    SPRING_DATASOURCE_HIKARI_MAXIMUM_POOL_SIZE: "4"
+    SPRING_DATASOURCE_HIKARI_MINIMUM_IDLE: "1"
+    DB_SCHEDULER_THREADS: "2"
+    DB_SCHEDULER_POLLING_INTERVAL_SECONDS: "60"
```
scottstewart-sl: Hmmm, this might slow the automated tests down a bit while they wait for access codes for example. If a build was very unlucky it could add about 50 seconds extra to each journey that waits for citizen access codes.

linusnorton: I've removed that change

scottstewart-sl (approved):

arun-hmcts (approved):

PR #2389, opened 2026-08-10
HDPI-7659: Fix missing case file documents caused by type-based filtering

sadmanrahh:
### Jira link

See [HDPI-7659](https://tools.hmcts.net/jira/browse/HDPI-7659)

### Change description

Resolve bug: GenApps Issued supporting document is not displayed in the case file view

### Testing done

Manual testing locally and in preview.

  * [ ] Yes
  * [X] No

### Checklist

### PCS checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/PCSCaseView.java
```diff
@@ -173,10 +181,106 @@ private SubmittedCase getSubmittedCase(long caseReference) {
         caseListView.setCaseFields(pcsCase);
         defendantResponseView.setCaseFields(pcsCase, pcsCaseEntity);
         featureFlagView.setCaseFields(pcsCase);
+        removeDocumentsAlreadyPresentInOtherCaseFields(pcsCase);
 
         return new SubmittedCase(pcsCase, pcsCaseEntity);
     }
 
+    private void removeDocumentsAlreadyPresentInOtherCaseFields(PCSCase pcsCase) {
```
scottstewart-sl: Can you move this new code into a separate service please?

Also I can see that the use of reflection would be an advantage in detecting new document fields added in the future, but I'm wondering about any performance implications, (given that this method is called frequently), while we are recursively traversing case data paths that will never have documents in them. I'm also not so keen on the package name and the "allDocuments" field names being string values.

We didn't get a chance to discuss an approach on this, but my initial thoughts were to either just explicitly specify the existing case data fields where documents might be added, (like the front end code partly does now), or to track each document (or document list) as it was added, so we could easily get that list of document IDs in one call and remove the duplicates from `allDocuments`

sadmanrahh: Sure, updated. I’ve moved the de-duplication into a separate service and replaced the reflection traversal with an explicit list of current case-data/tab fields that can contain documents.

scottstewart-sl (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/document/CaseFileDocumentDeduplicationService.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    private void addDocumentIdsFromRentArrears(RentArrearsSection rentArrears, Set<String> documentIds) {
+        if (rentArrears != null) {
+            addDocumentIds(rentArrears.getStatementDocuments(), documentIds);
+        }
+    }
+
+    private void addDocumentIdsFromNoticeServedDetails(
+        uk.gov.hmcts.reform.pcs.ccd.domain.NoticeServedDetails noticeServedDetails,
```
scottstewart-sl: Minor - can you import this class so you don't need the fully qualified name?

sadmanrahh: Addressed, thanks

libanAbdirahman1 (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/domain/genapp/GenAppDocument.java
```diff
@@ -1,18 +1,17 @@
...
 import lombok.Data;
 import lombok.NoArgsConstructor;
-import uk.gov.hmcts.ccd.sdk.type.Document;
 
 @Builder
 @Data
 @NoArgsConstructor
 @AllArgsConstructor
-public class DocumentWithId {
+public class GenAppDocument {
```
scottstewart-sl: What was the reason for this change? The front end has a corresponding type that matches `DocumentWithId` so this might have broken it.

The `document` field would have a filename field of its own anyway, so could you have just used that?

sadmanrahh: Good spot, thanks. I've reverted this back to the existing frontend-facing shape: submissionDocument uses DocumentWithId again and supporingDocuments uses List<Document>. The duplicate handling is now done inside GenAppsView before mapping, so the payload contract stays compatible while still avoiding duplicate GenApp submission.supporting docs.

gmmagruder (approved):

PR #2423, opened 2026-08-14
HDPI-8264: incorporate GroupAccess to NoC process

benouaer:
### Jira link
- See [HDPI-8264](https://tools.hmcts.net/jira/browse/HDPI-8264)

### Change description
- incorporate GroupAccess to NoC process

### Testing done
--

  * [x] Yes
  * [ ] No

### Checklist

### PCS checklist


src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/CaseAccessGroupsUtil.java
```diff
@@ -21,24 +29,48 @@
...
 
-    public static List<ListValue<CaseAccessGroup>> deriveCaseAccessGroups(Set<PartyEntity> parties) {
+    public static List<ListValue<CaseAccessGroup>> deriveCaseAccessGroups(List<ListValue<Party>> claimants,
+                                                                          List<ListValue<Party>> defendants,
+                                                                          Set<PartyEntity> partyEntities) {
         List<CaseAccessGroup> caseAccessGroups = new ArrayList<>();
+        Map<UUID, PartyEntity> partyEntitiesMap = partyEntities.stream()
+            .collect(Collectors.toMap(PartyEntity::getId, identity()));
+
+        requireNonNullElse(claimants, List.<ListValue<Party>>of())
```
andrewdami11: On a draft case there are often no claim parties yet, so `allClaimants` could be null and no claimant access group gets built, right? The older code fell back to the organisation party in `partyEntities` that `initialiseClaimant` creates for drafts. Is it worth putting that fallback back + testing for a draft case?

src/main/java/uk/gov/hmcts/reform/pcs/service/LegalRepresentativePartyLinkService.java
```diff
@@ -53,7 +49,7 @@ public void linkLegalRepresentativeToParty(long caseReference, String partyId,
 
         PartyEntity defendantPartyEntity = getDefendantPartyEntity(caseEntity, partyId);
 
-        unlinkExistingRepresentation(UUID.fromString(partyId));
+        unlinkExistingRepresentation(UUID.fromString(partyId), organisationId);
```
andrewdami11: `unlinkExistingRepresentation` now looks for an active link using the new organisation id. On a notice of change from org A to org B, that search looks for B, so A's link can stay active. The old code turned off the active link for that party, whatever the organisation was. Is it worth doing that again, and adding a test where the old and new org ids are different?"

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/GroupAccessType.java
```diff
@@ -134,11 +133,10 @@ private static Map<Key, GroupAccessType> buildIndex() {
         return Map.copyOf(index);
     }
 
-    /** Keyed lookup so selection does not depend on the order these constants are declared in. */
-    public static Optional<GroupAccessType> forProfileAndRole(String organisationProfileId, PartyRole partyRole) {
-        return partyRole == null
-            ? Optional.empty()
-            : Optional.ofNullable(BY_PROFILE_AND_ROLE.get(new Key(organisationProfileId, partyRole)));
+    public static String forProfileAndRole(String organisationProfileId, PartyRole partyRole, String orgId) {
+        return BY_PROFILE_AND_ROLE.get(new Key(organisationProfileId, partyRole))
```
andrewdami11: `forProfileAndRole` looks up a profile and then calls a method straight away. If the profile is missing or unknown, that becomes a null pointer error, no? 
The new `organisation_profile_id` column could also be null, with no data filled in for existing parties. Opening an old case could then fail when access groups are built. Is it worth handling a missing profile safely, or filling in the data?

src/main/resources/db/migration/V016__add_group_access_columns.sql
```diff
@@ -2,7 +2,7 @@
 
 -- The profile decides which access type an organisation gets, so it decides the capacity the case
 -- is stamped with.
-ALTER TABLE public.party ADD COLUMN organisation_profile_ids varchar(255)[];
+ALTER TABLE public.party ADD COLUMN organisation_profile_id varchar(64);
```
andrewdami11: This change looks like the amendment of an existing older migration? To prevent things from _breaking_, is it perhaps better to leave the old migration alone, and add a new one for the change?
You probably know better, but I am wary of our DB migrations versions fragility

src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/CaseAccessGroupsUtil.java
benouaer: no, after initialise claimant is called the claimant party entity is created, and that's how the org can access the case

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/GroupAccessType.java
benouaer: organisationProfileId cannot be null, otherwise we cannot determine its access
when the claimant part is initialised, the first thing that's added is org id and profile id
for defendants, we're filtering on defendants with legal rep orgs
when a legal rep org is created, their profile id is always added, and when we can't get it we throw an exception

src/main/resources/db/migration/V016__add_group_access_columns.sql
benouaer: you're right, I only amended it because I know it's in one the PRs this one is stacked on top off
I'll make sure this change is done in the original PR

src/main/java/uk/gov/hmcts/reform/pcs/service/LegalRepresentativePartyLinkService.java
benouaer: you're right, I'll amend the code to look for an active org for the party, and set that to inactive, rather than use the org id. that should be safe since only one org can be active for each party
although using the new org Id shouldn't return any active orgs because it's not linked yet

src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/CaseAccessGroupsUtil.java
andrewdami11: `initialiseClaimant` does create the party entity, but `deriveCaseAccessGroups` only walks `allClaimants`, and `PartiesView` leaves that null when there are no claims yet. How would the claimant group be built on a draft case, _before_ claim parties exist?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/CreatePossessionClaim.java
```diff
@@ -50,12 +49,7 @@ public void configureDecentralised(DecentralisedConfigBuilder<PCSCase, State, Us
                 .initialState(State.AWAITING_SUBMISSION_TO_HMCTS)
                 .showSummary()
                 .name("Make a claim")
-                // Create only: a case carries no CaseAccessGroups until it holds a claimant party,
-                // so the group access roles cannot reach the event that creates one.
-                .grant(Permission.CRUD, UserRole.PCS_SOLICITOR)
                 .grant(Permission.CRUD, UserRole.GA_CLAIMANT_SOLICITOR)
```
andrewdami11, line 62: Have perhaps mistakenly removed the only role that lets solicitors create the case before group access exists (i.e. PCS_SOLICITOR), or this was intentional?

src/test/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/DocumentAccessTest.java
```diff
@@ -9,18 +9,14 @@
...
 import static org.assertj.core.api.Assertions.assertThat;
 import static uk.gov.hmcts.reform.pcs.ccd.accesscontrol.UserRole.CITIZEN;
-import static uk.gov.hmcts.reform.pcs.ccd.accesscontrol.UserRole.CLAIMANT_SOLICITOR;
 import static uk.gov.hmcts.reform.pcs.ccd.accesscontrol.UserRole.DEFENDANT;
 import static uk.gov.hmcts.reform.pcs.ccd.accesscontrol.UserRole.DEFENDANT_SOLICITOR;
-import static uk.gov.hmcts.reform.pcs.ccd.accesscontrol.UserRole.PCS_SOLICITOR;
 
 class DocumentAccessTest {
 
     private static final UserRole[] CREATE_READ_ROLES = {
```
andrewdami11: `AccessGrants.documentAccess()` now gives `GA_CLAIMANT_SOLICITOR` create/read, but this test list no longer checks any claimant solicitor role. Is it worth adding `GA_CLAIMANT_SOLICITOR` here so the new grant stays covered?

andrewdami11, line 19: `AccessGrants.documentAccess()` now gives `GA_CLAIMANT_SOLICITOR` create/read, but this test list no longer checks any claimant solicitor role. Is it worth adding `GA_CLAIMANT_SOLICITOR` here so the new grant stays covered?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/CreatePossessionClaim.java
benouaer: this was intentional, they should be able to create the case with the claimant-solicitor group role

src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/CaseAccessGroupsUtil.java
benouaer: you're right, didn't notice that return statement when no claims are found
Arun has made a change in his PR to set a creator flag on the claimant party when it's created, so I'll use that to identify the claimant

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/GroupAccessType.java
andrewdami11: Thanks — happy to accept that for the create/link paths. I am resolving this thread.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/util/CaseAccessGroupsUtil.java
andrewdami11: Thanks — the `claim_creator` approach on the tip looks good. Resolving.

andrewdami11 (approved):

src/integrationTest/java/uk/gov/hmcts/reform/pcs/ccd/event/TestCaseGenerationIT.java
```diff
@@ -75,13 +77,21 @@ public class TestCaseGenerationIT extends AbstractPostgresContainerIT {
...
     @BeforeEach
     void setUp() {
         setUpAuthenticatedUser();
 
         FeeDetails feeDetails = FeeDetails.builder().code("FEE0001").feeAmount(new BigDecimal("123.45")).build();
         when(feeService.getFee(any(FeeType.class))).thenReturn(feeDetails);
+
+        when(organisationService.getOrganisationIdForCurrentUser()).thenReturn(ORGANISATION_ID);
+        when(organisationService.getOrgProfileIdForCurrentUser()).thenReturn("SOLICITOR_PROFILE");
+        when(organisationService.getOrganisationIdForCurrentUser()).thenReturn(ORGANISATION_ID);
```
toby-plunkett: I think these two lines are duplicated

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/AccessGrants.java
```diff
@@ -59,7 +62,7 @@ static SetMultimap<HasRole, Permission> documentAccess() {
         grants.putAll(PCS_SOLICITOR, Permission.CR);
         grants.putAll(CITIZEN, Permission.CR);
         grants.putAll(DEFENDANT, Permission.CR);
-        grants.putAll(CLAIMANT_SOLICITOR, Permission.CR);
+        grants.putAll(GA_CLAIMANT_SOLICITOR, Permission.CR);
```
toby-plunkett, line 70: I see we've added claimant and GA_claimant, but we only grant access to the GA claimant here. Do we need to also assign it to regular Claimant? Or is it assigned by default during creation

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/AccessProfile.java
```diff
@@ -22,6 +31,16 @@ public enum AccessProfile implements HasRole {
...
 
+    GA_CLAIMANT("claimant", CRU,
+             LOCAL_AUTHORITY_CLAIMANT_ACCESS,
+             REAL_ESTATE_ORG_CLAIMANT_ACCESS,
+             PROPERTY_CONSTRUCTION_ORG_CLAIMANT_ACCESS,
+             NOT_FOR_PROFIT_ORG_CLAIMANT_ACCESS,
+             CHARITY_ORG_CLAIMANT_ACCESS),
+    GA_CLAIMANT_SOLICITOR("claimant-solicitor", CRU, SOLICITOR_ORG_CLAIMANT_ACCESS),
+    GA_DEFENDANT_SOLICITOR("defendant-solicitor", CRU, SOLICITOR_ORG_DEFENDANT_ACCESS),
+    DUTY_ADVISOR_REQUEST("duty-advisor-request", Set.of(R), GroupAccessType.DUTY_ADVISOR_ACCESS),
```
toby-plunkett: minor, but Above we've used the access directly e.g SOLICITOR_... but here we used GroupAccessType.DUTY....

toby-plunkett (changes requested):
made it through a chunk, will finish off later :)

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/AccessProfile.java
benouaer: oversight, will remove

src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/AccessGrants.java
benouaer: I was only replacing the old solicitor role with the new one. but you're right, if this is granting access to the claimant, it should apply to both solicitor and LA claimant

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/TestCaseGeneration.java
```diff
@@ -69,7 +69,6 @@ void configure(DecentralisedConfigBuilder<PCSCase, State, UserRole> configBuilde
                 .initialState(AWAITING_SUBMISSION_TO_HMCTS)
                 .showSummary()
                 .name(EVENT_NAME)
-                // Same claimant capacities as createPossessionClaim, which this mirrors.
                 .grant(Permission.CRUD, UserRole.GA_CLAIMANT_SOLICITOR)
                 .grant(Permission.CRUD, UserRole.CLAIMANT)
                 // Temporary: keeps the tool usable on AAT until the group access flags are on.
```
toby-plunkett: very minor but do we need something to remind us to tackle this? Not sure if we mark with //TODO to flag in the IDE?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/CaseType.java
```diff
@@ -32,8 +32,8 @@ public class CaseType implements CCDConfig<PCSCase, State, AccessProfile> {
     static final AccessProfile[] PARTY_VISIBLE_TAB_ROLES = {
         AccessProfile.CITIZEN,
         AccessProfile.DEFENDANT,
+        AccessProfile.CLAIMANT,
         AccessProfile.PCS_SOLICITOR,
```
toby-plunkett, line 37: do we need to remove PCS_SOLICITOR now we have replaced it?

src/test/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/DocumentAccessTest.java
```diff
@@ -9,19 +9,17 @@
...
-        PCS_SOLICITOR,
         CITIZEN,
         DEFENDANT,
-        CLAIMANT_SOLICITOR,
-        DEFENDANT_SOLICITOR
+        DEFENDANT_SOLICITOR,
```
toby-plunkett, line 22: Does this role need to be removed with this update?

benouaer: no, I'm leaving all these roles to the cleanup ticket HDPI-7333

src/main/java/uk/gov/hmcts/reform/pcs/ccd/CaseType.java
benouaer: no, I'm leaving all these roles to the cleanup ticket HDPI-7333

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/TestCaseGeneration.java
benouaer: I didn't add that comment, I'll probably remove it
we're going to remove pcs_solicitor in the next ticket HDPI-7333

toby-plunkett (approved):

PR #2433, opened 2026-08-16
HDPI-8150: add notes and end dates to all owasp suppressions

arun-hmcts:
### Jira link

[HDPI-8150](https://tools.hmcts.net/jira/browse/HDPI-8150)

Follow-up to #2431, addressing Taleb's review comment: suppressions need end dates
and notes, so they get re-reviewed once they lapse and so the justification shows up
in the OWASP dependency-check report.

### Change description

Metadata only — `config/owasp/suppressions.xml`:

- Added `<notes>` to the 8 blocks that had none (only the Tomcat block had them).
  The notes text is what dependency-check renders against each suppressed finding in
  the report, so without it a reviewer can't see why something was suppressed.
- Added `until` to the 4 blocks that had none (the three false positives and Tomcat).
  Without an end date a suppression never lapses, so it's never re-reviewed.
  - Genuine false positives ("not affected") → `2027-02-15`.
  - Temporary suppressions (waiting on an upstream fix) keep their existing
    September/October dates; Tomcat gets `2026-09-15`.

**No behavioural change.** Every `packageUrl` and `<cve>` is untouched — all 9 rules
match master exactly, verified by parsing both files and comparing the
`(packageUrl, cve-set)` pairs rule by rule. Nothing is suppressed that wasn't before,
and nothing has been widened or narrowed.

### Testing done

- Parsed both the master and branch versions of `suppressions.xml` and diffed the
  rule sets programmatically: 9 rules each, identical `packageUrl` + CVE sets.
- Asserted every `<suppress>` block now carries both `<notes>` and `until` — 0 missing.
- XML well-formedness check passes.
- Relying on the PR build for the dependency-check gate itself; since no rule changed,
  the OWASP result should be unchanged from master.

### Security Vulnerability Assessment

- [x] Yes
- [ ] No

CVE IDs: no *new* suppressions are introduced by this PR. It documents and time-bounds
the ones already on master after #2431 — CVE-2026-54399, CVE-2026-54428, CVE-2025-7962,
CVE-2020-29582, CVE-2026-64607, CVE-2026-53914, CVE-2026-54512/54513/54514/54515,
CVE-2026-66299.

Reason for suppression: unchanged from #2431 — three are false positives (CVE reported
against a different artifact or an unreachable code path), the rest are temporary while
waiting for a fixed upstream release.

Mitigating factors: every suppression now has an expiry, so it fails the build again and
gets re-reviewed if the underlying dependency hasn't moved on by then.

### Feature flag assessment

**Feature flag:** Does this PR introduce or change user-facing behaviour that should be behind feature flag(s)?

- [ ] Yes — flag(s):
- [x] No — reason: build configuration only, no runtime behaviour.

### Checklist

### DIFF
--- a/config/owasp/suppressions.xml
+++ b/config/owasp/suppressions.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
   <!--Please add all the false positives under the below section-->
-  <suppress until="2027-02-15">
+  <suppress until="2026-09-15">
     <notes><![CDATA[
     HDPI-8150: false positive. CVE-2026-54399/54428 are against httpcore5 (5.x);
     this is httpcore 4.x, a separate artifact that is not affected. Re-review by
@@ -11,7 +11,7 @@
     <cve>CVE-2026-54399</cve>
     <cve>CVE-2026-54428</cve>
   </suppress>
-  <suppress until="2027-02-15">
+  <suppress until="2026-09-15">
     <notes><![CDATA[
     HDPI-8150: false positive. CVE-2025-7962 is reported against the Jakarta
     Activation spec API rather than the angus-activation implementation we
@@ -20,7 +20,7 @@
     <packageUrl regex="true">^pkg:maven/org\.eclipse\.angus/angus-activation@.*$</packageUrl>
     <cve>CVE-2025-7962</cve>
   </suppress>
-  <suppress until="2027-02-15">
+  <suppress until="2026-09-15">
     <notes><![CDATA[
     HDPI-8150: false positive. CVE-2020-29582 concerns kotlin-stdlib temp-file
     permissions; kotlin-stdlib is a transitive dependency only and PCS creates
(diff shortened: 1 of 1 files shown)


andrewdami11 (approved):

config/owasp/suppressions.xml
```diff
@@ -1,52 +1,89 @@
...
     <packageUrl regex="true">^pkg:maven/org\.apache\.httpcomponents\.core5/httpcore5(-h2)?@5\.([0-3](\..*)?|4(\.[0-2])?)$</packageUrl>
     <cve>CVE-2026-54399</cve>
     <cve>CVE-2026-54428</cve>
   </suppress>
   <suppress until="2026-09-15">
+    <notes><![CDATA[
+    HDPI-8270: temporary. Waiting on httpclient5 5.6.3+ to arrive transitively.
+    Remove once the resolved version is past the affected range in the
+    packageUrl above.
+    ]]></notes>
```
scottstewart-sl: Is the warning for this still being generated event though we're on 5.6.4 on master?
```
% gw dependencies
...
+--- org.apache.httpcomponents.client5:httpclient5:5.6.4 (*)
```

arun-hmcts: Good catch - resolves to 5.6.4 and the regex only covers up to 5.6.2, so it was suppressing nothing. Removed in 568dbc7b3.

scottstewart-sl (approved):

PR #2438, opened 2026-08-17
Hdpi 8262 Display Review Date information in Case Notes tab

sadmanrahh:
### Jira link

See [HDPI-8262](https://tools.hmcts.net/jira/browse/HDPI-8262)

### Change description

-Display date in case notes tab
-Save review date creator data and date created on in DB
-Ordering of review dates to show latest added first
 -Review date past or has been cancelled to also show

### Testing done

  * [ ] Yes
  * [X] No

### Checklist

### PCS checklist


src/main/resources/db/migration/V014__add_review_date_audit_fields.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE public.case_review_date
+  ADD COLUMN created_by VARCHAR(255),
```
guygrewal77: does the column size need to be 255 ? case_note has it set to 50.

arun-hmcts (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/AddCaseReviewDate.java
```diff
@@ -40,7 +40,7 @@ public void configureDecentralised(DecentralisedConfigBuilder<PCSCase, State, Us
                 .grant(Permission.CRUD, UserRole.CTSC_ADMIN, UserRole.HEARING_CENTRE_ADMIN, UserRole.WLU_ADMIN)
                 .grantHistoryOnly(JUDICIAL_HISTORY_ROLES)
                 .showSummary()
-                .showCondition(ShowConditions.featureFlagsEnabled(CASEWORKER_EVENTS, RELEASE_1_DOT_2))
+                .showCondition(ShowConditions.featureFlagsEnabled(CASEWORKER_EVENTS, RELEASE_1_DOT_3))
```
scottstewart-sl: Is this change intentional? This event will already be enabled in prod as 1.2 is now live, so this change would hide the event.

src/main/resources/db/migration/V014__add_review_date_audit_fields.sql
scottstewart-sl: You could go with `TEXT` to remove the need to specify a size

src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/AddCaseReviewDate.java
sadmanrahh: No, it was a mistake, thanks for pointing out, have reverted now

src/main/resources/db/migration/V014__add_review_date_audit_fields.sql
sadmanrahh: Sure, update to TEXT so we do not need to choose an arbitrary size.

libanAbdirahman1 (approved):

scottstewart-sl (approved):

src/main/java/uk/gov/hmcts/reform/pcs/ccd/view/CaseNoteView.java
```diff
@@ -2,18 +2,23 @@
...
+import java.util.Comparator;
 import java.util.List;
+import java.util.stream.IntStream;
 
 @Component
 public class CaseNoteView {
 
     public void setCaseFields(PCSCase pcsCase, PcsCaseEntity pcsCaseEntity) {
         setCaseNoteFields(pcsCase, pcsCaseEntity.getCaseNotes());
+        setReviewDateFields(pcsCase, pcsCaseEntity.getReviewDates());
```
scottstewart-sl: @sadmanrahh - no need to change now, but just looking at this in more detail I think the review dates should have their own *View class to populate the response. They are combined with the case notes in the ExUI tab, but that's just in the UI and for the data itself it would be slightly cleaner to keep them separate. I can move this code out in the subsequent PR if you agree.

PR #2156, opened 2026-07-10
Hpdi 4176 noc from master

m014389b:
### Jira link

See [HDPI-4176](https://tools.hmcts.net/jira/browse/HDPI-4176)

### Change description

Integrate with decentralised runtime NoC which allows for bespoke challenge questions, validation methods and submit methods to be implemented.

Due to a limitation in CCD around dynamic lists the challenge question answers are set to defendant1 to allow ccd config to generate, **they will not be used for validation**.  

XUI-Webapp requires the below config to be set. Specifically the "nocBaseUrl" which will be used as a redirect to PCS-API defined validation/submit methods via the "ccd decentralised runtime nocController" 

`DECENTRALISED_CASE_TYPE_CONFIG={
    "PCS": {
      "nocBaseUrl": "https://pcs-api.service.core-compute.internal//ccd/noc"
    }
}`

PCSNoticeOfChange defines the validation and submit methods, invoked via the nocBaseUrl. This is where NoC challenge question validation will be performed, followed by the NoC submission logic linking a solicitor to a party 

Remove test support controller - linkDefendantSolicitorToParty as no longer required now have NoC 

AAC manage case requirements:

* Added logic to include organisationPolicy.OrgPolicyCaseAssignedRole of type "DEFENDANT_SOLICITOR" to each defendant party.

* A Noc Event available only to the "caseworker-caa"

### Testing done

Manual testing on preview with a solicitor with an organisation performing NoC

### Security Vulnerability Assessment ###

**CVE Suppression:** Are there any CVEs present in the codebase (either newly introduced or pre-existing) that are being intentionally suppressed or ignored by this commit?
  * [ ] Yes
  * [ ] No

### Checklist

- [ ] commit messages are meaningful and follow good commit message guidelines
- [ ] README and other documentation has been updated / added (if needed)
- [ ] tests have been updated / new tests has been added (if needed)
- [ ] Does this PR introduce a breaking change

PR #2439, opened 2026-08-17
HDPI-7052: make the group access config branch additive and mergeable

arun-hmcts:
### Jira link

[HDPI-7052](https://tools.hmcts.net/jira/browse/HDPI-7052)

### Change description

Declares group access for PCS, so everyone in an organisation sees that organisation's cases without
per-case sharing. Declaration only — deriving the id onto a case is
[#2179](https://github.com/hmcts/pcs-api/pull/2179). Supersedes #1978.

| Access type | Organisation profiles | Group role |
|---|---|---|
| `prof-org-claimant-access` | LOCALAUTH, OTHER_REALT, OTHER_PROP, OTHER_NFP, OTHER_CHARITY | `claimant` |
| `solicitor-org-claimant-access` | SOLICITOR | `claimant-solicitor` |
| `solicitor-org-defendant-access` | SOLICITOR | `defendant-solicitor` |
| `duty-advisor-access` | SOLICITOR | `duty-advisor-request` |

Plus the matching `UserRole` / `AccessProfile` entries, case-type and event grants,
`HighLevelDataSetupApp` role creation, and `PCSCase.caseAccessGroups`. SDK pinned to 6.33.0 for
`CaseAccessGroup`.

`caseworker-pcs-solicitor` is kept everywhere it exists today — no existing access changes.

### Preview no longer runs its own copies of shared components

Group access is now enabled in AAT — `ENABLE_CASE_GROUP_ACCESS` on the definition store and
`ENABLE_CASE_GROUP_ACCESS_FILTERING` on the data store
([cnp-flux-config#47000](https://github.com/hmcts/cnp-flux-config/pull/47000)) — and RAS carries the
four PCS roles since
[am-role-assignment-service#2781](https://github.com/hmcts/am-role-assignment-service/pull/2781).

Before [#2156](https://github.com/hmcts/pcs-api/pull/2156) (NoC) a preview used AAT for all of this.
That PR added per-PR copies and pointed several components at a per-PR role-assignment service —
but `ccd.ras.enabled` has been `false` since the chart was first added and was never once true on
master, so those pointers have been aimed at a service that is never deployed. This restores the
earlier arrangement.

| Removed from the preview chart | Why |
|---|---|
| `am-org-role-mapping-service` | wrote only to `${SERVICE_NAME}-am-role-assignment-service`, which is never deployed |
| `xui-mo-webapp` | AAT's Manage Organisation covers it once the access types this PR adds are on AAT |
| `aac-manage-case-assignment` | shared platform service; the xui-webapp subchart default resolves to AAT's instance |
| `pr-N-role-assignment` database | referenced by nothing at all |
| `pr-N-org-role-mapping` database | referenced only by the ORM above |

`xui-webapp`'s `SERVICES_ROLE_ASSIGNMENT_API` / `_MAPPING_API` now name AAT explicitly rather than a
per-PR service. The S2S authorised-service lists keep `aac_manage_case_assignment`, since AAT's
instance calls the preview's stores under the same name.

The chart now deploys exactly the pre-NoC set — `global`, `java`, `postgresql`, `ccd`,
`xui-webapp`, `ccd-case-document-am-api` and three databases — plus the two
`ENABLE_CASE_GROUP_ACCESS*` flags, since the preview runs its own CCD stores. Everything still
per-PR is the CCD and document stack this branch changes; every shared dependency comes from AAT.

**104 lines removed, 4 added.** Three fewer pods and two fewer databases per preview, and no user
seeding.

Trade-off worth calling out: AAT's aac reads AAT's data store, so NoC journeys can no longer be
exercised against a preview's own cases.

### Testing done

image

Generated definition compared against master (`d8d2c7df1`) as row sets — inserting roles renumbers
display order, so a text diff is unreadable:

| File | removed | added |
|---|---|---|
| `AccessType.json` | 0 | 8 |
| `AccessTypeRole.json` | 0 | 8 |
| `AuthorisationCaseType.json` | 0 | 4 |
| `CaseField.json` | 0 | 1 |
| `RoleToAccessProfiles.json` | 0 | 4 |
| **total** | **0** | **25** |

**Nothing removed anywhere** — no existing role loses a permission.

The four new profiles get case-type access and nothing else: no `AuthorisationCaseField`,
`CaseTypeTab`, `AuthorisationCaseEvent` or `AuthorisationCaseState` rows, since a43a8eb9 dropped them
from the party-visible tabs. So this branch cannot change what any existing user sees; the grants
that make organisation cases readable land with the deriving half.

With the preview on AAT's RAS, staff and judicial users hold their real roles with nothing seeded.

Unit tests: `ClaimantAccessTest`, `RoleToAccessProfilesTest`, `CaseTypeTest`, `OrganisationServiceTest`.

  * [ ] Yes
  * [x] No

`config/owasp/suppressions.xml` is byte-identical to master.

### Checklist

### PCS checklist

- [ ] New functional test classes carry the gating annotations (HDPI-8084) — none added
- [ ] Infrastructure / chart / suppression changes have a second pair of eyes from the service admins (HDPI-8153) — chart version bump, and the preview now uses AAT's RAS


bin/create-xlsx.sh
```diff
@@ -20,6 +20,7 @@ for case_dir in "$run_dir"/build/definitions/*/; do
 
   ccd_definition_file="CCD_Definition_${case_type}_${env}.xlsx"
 
+#comment out two lines below to run locally without Azure login
```
scottstewart-sl: This is fine for now

When I have time, I was planning on adding a check for the `CI` env var which should only be set on Jenkins

charts/pcs-api/values.ccd.preview.template.yaml
```diff
@@ -86,6 +86,7 @@ ccd:
         DEFINITION_STORE_DB_MAX_POOL_SIZE: 2
         ELASTIC_SEARCH_INDEX_SHARDS: 1
         ELASTIC_SEARCH_INDEX_SHARDS_REPLICAS: 0
+        ENABLE_CASE_GROUP_ACCESS: true
```
scottstewart-sl, line 87: This suggests that `ENABLE_CASE_GROUP_ACCESS` is not enabled by default in ccd-definition-store-api. Will it be enabled for the various environments in Flux config?

charts/pcs-api/values.ccd.preview.template.yaml
benouaer: I think if we're removing role-assignment-service, we also need to remove aac-manage-case-assignment, am-org-role-mapping-service and xui-mo-webapp

bin/create-xlsx.sh
arun-hmcts: Good shout on the `CI` env var check — that's cleaner than the comment I'd added. I've dropped this file from the PR though, it was unrelated noise that crept in, so `create-xlsx.sh` is back to master now.

charts/pcs-api/values.ccd.preview.template.yaml
arun-hmcts: Yeah, it's off by default in the definition store. Right now it's only on in AAT ([cnp-flux-config#47000](https://github.com/hmcts/cnp-flux-config/pull/47000)) — prod, demo, ithc and perftest don't set it, so they're still false. The preview needs its own flag because it runs its own CCD stores rather than pointing at AAT's. The access types will import into the other environments and just sit inert until Flux switches it on there.

charts/pcs-api/values.ccd.preview.template.yaml
arun-hmcts: Agreed — done in fafaa225b.

To be precise about what was there: the chart doesn't deploy a role-assignment-service at all, and that was already true on master (`ras.enabled: false`). What was left behind was three things still pointing at a per-PR one that never gets deployed — `xui-webapp` (204/205), `am-org-role-mapping-service` (306) and `aac-manage-case-assignment` (339). So the per-PR ORM was writing role assignments into the void, and aac's role-assignment calls went nowhere.

Removed `am-org-role-mapping-service` and `xui-mo-webapp`, plus the `pr-N-role-assignment` and `pr-N-org-role-mapping` databases — the first was referenced by nothing at all. Repointed xui-webapp and aac at AAT.

Kept `aac` and `xui-webapp` themselves, since both read the preview's own definition store and case data — AAT's would show AAT cases rather than the preview's. It was only their role-assignment URLs that were wrong.

Net: 3 insertions, 63 deletions, two fewer pods and two fewer databases per preview.

arun-hmcts: Update — aac is gone too in `c6094b3c4`, so that's all three you listed; the chart is back to the pre-NoC set, 104 lines removed.

benouaer (approved):

gmmagruder (approved):

linusnorton (approved):

scottstewart-sl (approved):

PR #2459, opened 2026-08-20
HDPI-7770: review case flag request changes

sadmanrahh:
### Jira link

See [HDPI-7770](https://tools.hmcts.net/jira/browse/HDPI-7770)

### Change description

Implements the Review Case Flag Request Work Allocation task

### Testing done

  * [ ] Yes
  * [X] No

### Checklist

### PCS checklist


src/main/resources/dmn/wa-task-completion-possession_service.dmn
```diff
@@ -99,7 +99,7 @@
       </rule>
       <rule id="DecisionRule_0p08ndn">
         <inputEntry id="UnaryTests_1ml8wgu">
-          <text>"Approve or reject case flag"</text>
+          <text>"Review Support Request"</text>
```
andrewdami11: This Auto rule completes `ReviewCaseFlagRequest` on `\"Review Support Request\"`, but PCS Case Events today are `amendFlags` (`Manage case flags`) and `createFlags` (`Create case flags`). There is no matching event id or name `\"Review Support Request\"` in the CCD config. Is it worth confirming the real published event id (for example `amendFlags`) and aligning this Auto rule to that, so the task actually closes?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -82,6 +83,18 @@ public void saveReasonableAdjustmentFlags(PartyEntity partyEntity, Flags incomin
             return;
         }
 
+        List<String> requestedFlags = reasonableAdjustmentDetails.stream()
+            .map(ListValue::getValue)
+            .filter(CaseFlagService::isCaseFlagRequested)
+            .map(FlagDetail::getName)
+            .toList();
+
+        if (!CollectionUtils.isEmpty(requestedFlags)) {
```
andrewdami11: Task creation for `REVIEW_CASE_FLAG_REQUEST` only runs in `saveReasonableAdjustmentFlags` (CUI RA). `mergePartyFlags` / `mergeCaseFlags` (used by Manage/Create case flags) still do not create this task, while the initiation DMN note still mentions external professional users adding flags via manage case flags. Is it worth clarifying whether that path is out of scope for HDPI-7770, or adding the same Requested-status trigger there?

src/test/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagServiceTest.java
```diff
@@ -223,6 +224,38 @@ void shouldIgnoreSuppliedFlagsThatAreNotReasonableAdjustments() {
         verify(camundaService).createTask(CASE_REFERENCE, TaskType.REVIEW_CASE_FLAG, "description");
     }
 
+    @Test
+    void shouldCreateOneReviewCaseFlagRequestTaskForMultipleRequestedReasonableAdjustments() {
```
andrewdami11: The new test covers multiple Requested flags well, and Active-only paths are already covered. Is it worth one more test with both `Requested` and `Active` RA flags in one payload, so we lock in both `REVIEW_CASE_FLAG_REQUEST` and `REVIEW_CASE_FLAG` being created once each?

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/workallocation/TaskDescriptionService.java
```diff
@@ -158,6 +158,16 @@ public String createReviewCaseFlagDescription(long caseReference, List<String> f
         return renderTemplate(templateName, context);
     }
 
+    public String createReviewCaseFlagRequestDescription(long caseReference, List<String> flags) {
+        Map<String, Object> context = Map.of(
+            "caseReference", caseReference,
+            "flags", flags
+        );
+
+        String templateName = "review-case-flag-request";
```
andrewdami11: `createReviewCaseFlagRequestDescription` loads template `review-case-flag-request`, but there is no `src/main/resources/templates/workallocation/review-case-flag-request.peb` (only `review-case-flag.peb` exists). The unit test _stubs_ Pebble, so it would not catch a missing file. Is it worth adding the `.peb` file (aligned with `review-case-flag.peb`)?

sadmanrahh: Addressed. Added review-case-flag-request.peb aligned with the existing review case flag template.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
sadmanrahh: Good spot. I’ve added the same Requested status trigger to the case and party flag merge paths, so manage/create case flags can also create the ReviewCaseFlagRequest task.

src/main/resources/dmn/wa-task-completion-possession_service.dmn
sadmanrahh: Good spot. I checked the HDPI-7656 branch and the Review Support Request event is being implemented there as the CCD event ID reviewSupportRequest. I've aligned this completion rule to that event ID rather than amendFlags / createFlags, because the task should complete when the support request is reviewed, not when flags are created or managed.

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
```diff
@@ -136,6 +153,30 @@ public void mergePartyFlags(List<ListValue<Party>> incomingParties, Set<PartyEnt
                 }
             }
         }
+
+        createReviewCaseFlagRequestTask(caseReference, requestedFlags);
```
scottstewart-sl: Are the flags coming into this method actually flag _requests_ ? I'm not very familiar with the flag logic here but the comment on `saveReasonableAdjustmentFlags(...)` suggests that this `mergePartyFlags(...)` method is used for caseworker flags

sadmanrahh: Good point. mergePartyFlags() is used by the internal Manage/ Create case flags flow, so treating those incoming party flags as support requests is too broad. I've removed the ReviewCaseFlagRequest task creation from this method. The review-request task creation remains on the actual requested flag flow instead.

scottstewart-sl (approved):

featureFlags/flags.json
```diff
@@ -6,6 +6,6 @@
     "release-1.2-enabled": true,
     "release-1.3-enabled": true,
     "wales-make-a-claim-enabled": true,
-    "caseworker-wa-enabled": false
+    "caseworker-wa-enabled": true
```
andrewdami11: Is this intended to be changed to `true`? Just double checking

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
andrewdami11: As per the latest commit though, this only fires for CUI RA and case-level `mergeCaseFlags`, not `mergePartyFlags`. Is party-level Requested via Manage/Create out of scope for HDPI-7770? If yes I’ll resolve, otherwise we may still need something on that path, I think

featureFlags/flags.json
sadmanrahh: Yes just temporarily enabled to test caseworker WA, will change it back to false before merging

src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseFlagService.java
sadmanrahh: Yes, party-level Requested via Manage/Create is out of scope for HDPI-7770. I removed the mergePartyFlags trigger after Scott’s review because that path is the internal caseworker Manage/Create flags flow, and the party-level support review flow is covered by HDPI-7656. This PR now keeps the WA task aligned with the HDPI-7656 event/completion path without duplicating that wider scope.

andrewdami11 (approved):

featureFlags/flags.json
scottstewart-sl: This file is only for local dev anyway, so you don't need to commit it to control feature flags in preview.

PR #2471, opened 2026-08-21
Fix functional test create-case retry handling

gergelykisshmcts:
### Jira link

Personal contribution no jira link

### Change description

createCase only retried unsuccessful HTTP status codes. A transport-level failure (TLS/socket) was not retried, and SerenityRest turned it into a ResponseStub while recording a failed step on the Serenity event bus. Serenity then skips subsequent @Step calls and returns Mockito mocks from them, so a create-case call that was successfully retried still failed the test and surfacing as the misleading claimantPaymentRef NPE seen on master.

- retry both unsuccessful responses and request exceptions
- keep the final failure as the cause of the thrown IllegalStateException
- include the status code and response body when a call returns non-201
- issue the request with plain RestAssured so a recovered retry no longer records a failed step and poisons later steps
- assert the first fee payment reference is non-null before dereferencing it

### Testing done

Local test runs

  * [ ] Yes
  * [X] No

### Checklist

### PCS checklist


scottstewart-sl (approved):

tvr-hmcts (approved):
Cool. :)

Some thoughts ... perhaps beyond the fix though.

Somethings can not be retried so should they be checked for and just skip out earlier on?  400 / 401's etc

Also there is no back off / wait etc period before hitting again in that loop.  Would that help at all in some circumstances?

marianadpereira (approved):

PR #2492, opened 2026-08-25
HDPI-7052: set devMode memory limits — staging/preview pods are OOMKilled at the chart's 1Gi default

arun-hmcts:
Sets the devMode memory keys (used by AAT staging and PR previews) so those pods get a 2Gi limit instead of the java chart's 1Gi default, which the E2E runs have been OOMKilled against.


srinijg (approved):
LGTM

scottstewart-sl (approved):

PR #2493, opened 2026-08-25
stop throwing exception when org not found for non prof users

benouaer:
### Jira link

See [PROJ-XXXXXX](https://tools.hmcts.net/jira/browse/PROJ-XXXXXX)

  * [ ] Yes
  * [ ] No

### Checklist

### PCS checklist

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/pcs/reference/service/OrganisationDetailsService.java
+++ b/src/main/java/uk/gov/hmcts/reform/pcs/reference/service/OrganisationDetailsService.java
@@ -1,12 +1,13 @@
 package uk.gov.hmcts.reform.pcs.reference.service;
 
+import static java.util.Objects.nonNull;
+
 import feign.FeignException;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Qualifier;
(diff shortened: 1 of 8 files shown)


arun-hmcts (approved):

src/main/java/uk/gov/hmcts/reform/pcs/reference/service/OrganisationDetailsService.java
```diff
@@ -51,11 +52,11 @@ public OrganisationDetailsResponse getOrganisationDetails(String userId) {
         } catch (FeignException ex) {
             log.error("Feign error retrieving organisation details for userId: {}. Status: {}, Message: {}",
                 userId, ex.status(), ex.getMessage(), ex);
-            throw new OrganisationDetailsException("Failed to retrieve organisation details", ex);
+            return null;
```
scottstewart-sl: This is fine, but it might be nice, (maybe in a later PR), to distinguish between a 404 response and other errors, and still throw the exception if it wasn't a 404. Then this method could return an `Optional` which would be a bit more convenient to work with in methods like `getOrganisationName` below.

scottstewart-sl (approved):

PR #2495, opened 2026-08-26
HDPI-6033: return a validation error, not a 500, when a make-an-application submit names a party that isn't on the case

arun-hmcts:
### Jira link

See [HDPI-6033](https://tools.hmcts.net/jira/browse/HDPI-6033)

### Change description

A make-an-application submit whose `currentRepresentedPartyId` did not resolve on the case threw an unhandled `PartyNotFoundException` → HTTP 500 (CCD reports this to callers as a 502, indistinguishable from an outage). The id is client-supplied case data, so this catches it in the submit handler and returns the handler's existing `errorResponse` validation message instead. No change to the party lookup, guard, or queries.

### Testing done

Renamed the unit test that asserted the old throw to assert the new validation response (`SubmitEventHandlerTest`); it passes and `errorResponse` is verified. Also reproduced the original 500 as a functional test against a preview before the fix.

  * [ ] Yes
  * [x] No

### Checklist

### PCS checklist

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/genapp/SubmitEventHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/genapp/SubmitEventHandler.java
@@ -77,10 +77,7 @@ public class SubmitEventHandler implements Submit<PCSCase, State> {
         try {
             applicantParty = getApplicantParty(caseReference, caseData);
         } catch (PartyNotFoundException | IllegalArgumentException ex) {
-            // currentRepresentedPartyId is client-supplied case data - the server sets it from a
-            // case+org scoped list at start, so a value that no longer resolves is bad input, not
-            // a server fault. Return a validation error rather than a 500 (which CCD reports as a
-            // 502, indistinguishable from an outage).
+            // client-supplied party id that no longer resolves is bad input, not a server fault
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/pcs/ccd/event/genapp/SubmitEventHandler.java
```diff
@@ -70,7 +72,16 @@ public SubmitResponse<State> submit(EventPayload<PCSCase, State> eventPayload) {
...
         PcsCaseEntity pcsCaseEntity = pcsCaseService.loadCase(caseReference);
-        PartyEntity applicantParty = getApplicantParty(caseReference, caseData);
+
+        PartyEntity applicantParty;
+        try {
+            applicantParty = getApplicantParty(caseReference, caseData);
+        } catch (PartyNotFoundException | IllegalArgumentException ex) {
+            // client-supplied party id that no longer resolves is bad input, not a server fault
+            log.warn("Applicant party could not be resolved on genapp submit for case {}: {}",
+                caseReference, ex.getMessage());
```
scottstewart-sl: It might be useful to have the party ID in the log message as well, in case it helps tracking down where they are coming from

scottstewart-sl (approved):

PR #2509, opened 2026-08-27
HDPI-7052: hide group-access types from ManageOrg on the staging case type

arun-hmcts:
### Change description

The staging definition (`CASE_TYPE_SUFFIX=staging`) duplicates every AccessType row of the main PCS case type, so ManageOrg's **Additional types of access** screen shows each checkbox twice. This emits `Display=N` on the staging rows; ManageOrg already filters on that flag, so the duplicates disappear.

Safe: user ticks are keyed without case type, and ORM never reads Display — `PCS-staging` role derivation is unchanged.

Also fixes NoC back to a previous organisation leaving the party with no active organisation link. `OrganisationEntity.addParty` skipped creating a new link whenever *any* row existed for the party — including inactive history rows — so an A → B → A notice-of-change sequence deactivated the current link but never created the replacement, and the party ended up with no active organisation (repro: case 1787845845527279 on AAT). The guard now only skips on an active link, comparing party ids by value rather than reference.

### Testing done

`generateCCDConfig`: no suffix → `Display=Yes`; `CASE_TYPE_SUFFIX=staging` → `Display=No`.

NoC re-link: new unit tests in `OrganisationEntityTest` cover the A → B → A re-link (fresh active row created alongside the historic inactive one) and value-based id dedup; verified against AAT data that the broken case's party had zero active links after the third NoC.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/AccessProfile.java
+++ b/src/main/java/uk/gov/hmcts/reform/pcs/ccd/accesscontrol/AccessProfile.java
@@ -19,6 +19,7 @@ import lombok.Getter;
 import uk.gov.hmcts.ccd.sdk.api.CCDAccessGroup;
 import uk.gov.hmcts.ccd.sdk.api.HasRole;
 import uk.gov.hmcts.ccd.sdk.api.Permission;
+import uk.gov.hmcts.reform.pcs.ccd.CaseType;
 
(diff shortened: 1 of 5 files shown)


arun-hmcts:
Superseded by #2577 — same change, the branch was renamed to HDPI-8685-noc-relink-previous-org and GitHub closed this PR during the rename.

PR #2514, opened 2026-08-27
HDPI-8685: queue a search reindex when NoC changes case representation

arun-hmcts:
### Jira link

[HDPI-8685](https://tools.hmcts.net/jira/browse/HDPI-8685)

### Change description

Fixes **Issue 1**: after a Notice of Change the incoming solicitor can open the case but it never appears in their case list, and the outgoing org keeps a stale row (404 on click).

`CaseAccessGroups` is derived at read time and only reaches Elasticsearch when the case is re-indexed — which happens via a trigger on `ccd.case_data` writes. The NoC apply task updates `claim_party_organisation` **without writing the case row**, so nothing is queued and the index keeps the pre-NoC groups (verified on AAT: case `1787849985141941` — direct GET 200 as the new rep, searchCases total 0; `last_modified` unchanged in DB and ES).

Fix: after the representation link is persisted, queue the case on `ccd.es_queue` (same insert/conflict clause as the SDK's own trigger, without touching the case row). The existing indexer refetches, derives fresh groups, and rewrites the document — the new org's list shows the case and the outgoing org's stale row disappears in the same pass. Runs inside the task's retry envelope so link + reindex retry together.

### Notes

- Repeat NoC (A→B→C, and A→B→A) verified safe: `unlinkExistingRepresentation` ends the previous link and derivation filters to the active link, so the index converges to the current org each hop.
- Cases NoC'd before this fix stay stale until re-queued — backfill via the SDK's existing `CaseReindexingService.enqueueCasesModifiedSince(...)` after deploy.
- **Issues 2/3** on the ticket (regain-access / missing `claim_party_organisation` row on second NoC) look like a separate persistence defect in the relink path — not addressed here.
- SDK counterpart raised: [dtsse-ccd-config-generator#1077](https://github.com/hmcts/dtsse-ccd-config-generator/pull/1077) adds `CaseReindexingService.reindexCase(caseRef)` — once released, `CaseReindexService` here is replaced by that call.

### Testing done

Unit tests: reindex queued after a successful link; not queued when the link fails (failure propagates for retry). `compileJava`, `checkstyleMain/Test`, `NocAccessChangeTaskComponentTest` green.

AAT verification after merge: repeat the ticket's repro — after NoC as `pcs-org1-solicitor3@test.com`, the case appears in their case list within seconds, and the previous org's list drops it.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseReindexService.java
+++ b/src/main/java/uk/gov/hmcts/reform/pcs/ccd/service/CaseReindexService.java
@@ -1,0 +1,43 @@
+package uk.gov.hmcts.reform.pcs.ccd.service;
+
+import java.util.Map;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Service;
+
+/**
+ * Queues a single case for Elasticsearch reindexing.
+ *
+ * <p>CaseAccessGroups is derived at read time, so it only reaches the search index when the
+ * indexer refetches the case — which normally happens because every CCD event writes
+ * {@code ccd.case_data} and a trigger enqueues the case. Writes that change derivation inputs
+ * outside an event (e.g. a Notice of Change updating {@code claim_party_organisation}) must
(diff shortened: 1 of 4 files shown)


arun-hmcts:
Superseded by the SDK-consuming version of the same fix — see the replacement PR referencing dtsse-ccd-config-generator#1077.

PR #2515, opened 2026-08-27
HDPI-8685: record a noticeOfChangeApplied event so the incoming solicitor's case list picks up the case

arun-hmcts:
### Jira link
https://tools.hmcts.net/jira/browse/HDPI-8685

### Problem
After a Notice of Change the new solicitor's firm doesn't see the case in their case list (the old firm keeps a stale row).

### Cause
The case list is served from Elasticsearch, which is rebuilt only from the snapshot saved by the last case event. NoC updates our tables but raises no event, so the index keeps the old firm's `CaseAccessGroups`. Re-queueing the case for indexing (SDK #1077) doesn't help – it re-emits the same old snapshot.

### Fix
After the NoC link is saved, submit a hidden system event `noticeOfChangeApplied`. That stores a fresh snapshot, so the index picks up the new firm and drops the old one. Side effect: "Notice of change applied" now shows in the case history (recorded by the system user). No SDK change.

Cases NoC'd before this needs the event submitted once each.

### Also
- OWASP suppression for CVE-2026-39882/39883 on `opentelemetry-semconv` (OpenTelemetry-Go issues, generic CPE match).

### Tested
- Unit tests for the task and the event service.
- Preview: NoC as `eviction-solicitor-user1` on `1787991572822903` – case appears for both users in that firm, unchanged for the claimant, hidden from an unrelated firm; history shows the event.


src/main/java/uk/gov/hmcts/reform/pcs/noc/event/NoticeOfChangeApplied.java
```diff
@@ -0,0 +1,57 @@
...
+        UserRole.GA_DEFENDANT_SOLICITOR,
+        UserRole.CITIZEN,
+        UserRole.PCS_CASE_WORKER
+    };
+
+    @Override
+    public void configureDecentralised(DecentralisedConfigBuilder<PCSCase, State, UserRole> configBuilder) {
+        configBuilder
+            .decentralisedEvent(noticeOfChangeApplied.name(), this::submit)
+            .forAllStates()
```
benouaer: We should restrict this only to states where NoC is applicable. Would that be CaseIssued?

PR #2525, opened 2026-08-28
HDPI-8701: give group-access roles Request/Manage support, the Support tab and Make an application

arun-hmcts:
### Jira link

See [HDPI-8701](https://tools.hmcts.net/jira/browse/HDPI-8701)

### Change description

Users who reach a case through group access (the org-wide `claimant`, `claimant-solicitor` and `defendant-solicitor` roles) were missing three things their per-case-role colleagues already had. This PR closes those gaps so both kinds of user get the same options on a case.

- **Request support / Manage support** – added the three group-access roles to `ExternalCaseFlagRoles`, so both events appear under Next step for them (and the flag fields they write to are accessible).
- **Support tab** – the tab that shows the requested support was only generated for `caseworker-pcs-solicitor`, citizen and defendant. Added the three group-access roles so they can see what they've requested. The per-case roles stay off it, as before.
- **Make an application (claimant)** – `claimantMakeAnApplication` was granted to `[CLAIMANTSOLICITOR]` only. Added `claimant-solicitor` and `claimant`, matching what the defendant side already does. This also unlocks the application fields for those users.

Checked by generating the CCD definition and comparing each group-access role against its per-case counterpart across events, states, fields and tabs – no differences remain on the defendant side, and on the claimant side the only remaining one is the `CLOSED` state (no claimant-side role can read a closed case today, including the legacy one), which is being raised separately.

Also bumps an e2e timeout for case file load and registers the `confirmStatusForFlag` e2e action.

### Testing done

- Unit: `RequestSupportTest`, `ManageSupportTest`, `ExternalCaseFlagAccessTest` (new role grants); `CaseTypeTest` (Support tab roles); new `ClaimantMakeAnApplicationTest` (claimant grants, no defendant grants).
- Definition: regenerated and diffed group-access vs per-case roles – 0 event / state / field gaps on both sides after the change.
- PR preview runs the `@CC` e2e suite, which includes `caseSupport.spec.ts`.

  * [ ] Yes
  * [x] No

### Checklist

### PCS checklist


lovelin-07 (approved):
LGTM

benouaer (approved):

scottstewart-sl (approved):

